PR #314, opened 2013-04-08
Animated text does not always stack vertically correctly.

Quintinon:
Fix for Animated text not stacking properly vertically.
Change will also merge text of the form "+123" since some servers use this for healing.

I think the style followed the code conventions properly, but it might be a little redundant on calculations, maybe they should be moved to a function inside of the Animated Text class?


Quintinon:
I've discarded these changes as my repo was too old for me to bother merging. I'm closing this pull request. I may in the future redo this fix if I run across the issue again. (it seems to not be very apparent for most people)

PR #595, opened 2014-11-18
Added Exploration Mode (Crash Fix)

LordHepipud:
- Exploration mode:
  By default OTClient is drawing all tiles which are send by the server to the minimap view. Some server admins might not want this and will receive a working patch for this. Simply add EXPLORE_MODE to build dependencies to make it work. While still all tiles are received from the server, only the visible ones on the same floor as the player including tiles inside the view range will be drawn. All other tiles will be added to a caching system and drawn later in case the player walks up or down a floor. To prevent the Client from freezing or slowing done after a long play time, the cache will be flushed of all tiles which have been drawn already or which are far out from our view range

Improvements:
- Names of caching variables changed (added m_ before)
- Some code cleanup

Fixes:
- Crash bug while exploring new areas in some cases

### DIFF
--- a/modules/corelib/ui/uiscrollbar.lua
+++ b/modules/corelib/ui/uiscrollbar.lua
@@ -28,7 +28,7 @@ local function calcValues(self)
     proportion = math.min(math.max(self.step, 1), range)/range
   end
 
-  local px = math.max(proportion * pxrange, 10)
+  local px = math.max(proportion * pxrange, 6)
   px = px - px % 2 + 1
 
   local offset = 0
--- a/src/client/minimap.cpp
+++ b/src/client/minimap.cpp
@@ -24,6 +24,9 @@
 #include "minimap.h"
 #include "tile.h"
 
+//PhiadariaSoft Mod: Minimap drawing fix to prevent drwaing multiple floors
+#include "game.h"
+
 #include <framework/graphics/image.h>
 #include <framework/graphics/texture.h>
 #include <framework/graphics/painter.h>
@@ -96,6 +99,9 @@ void Minimap::clean()
 {
     for(int i=0;i<=Otc::MAX_Z;++i)
         m_tileBlocks[i].clear();
+        
+    m_minimapFloorCache.clear();
+	m_minimapDrawCache.clear();
 }
 
 void Minimap::draw(const Rect& screenRect, const Position& mapCenter, float scale, const Color& color)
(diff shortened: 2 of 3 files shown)


diegorodriguesvieira:
Thanks a lot dude! =)

src/client/minimap.cpp
```diff
@@ -203,12 +209,100 @@ void Minimap::updateTile(const Position& pos, const TilePtr& tile)
         minimapTile.speed = std::min<int>((int)std::ceil(tile->getGroundSpeed() / 10.0f), 255);
     }
 
-    if(minimapTile != MinimapTile()) {
-        MinimapBlock& block = getBlock(pos);
-        Point offsetPos = getBlockOffset(Point(pos.x, pos.y));
-        block.updateTile(pos.x - offsetPos.x, pos.y - offsetPos.y, minimapTile);
-        block.justSaw();
-    }
+#ifndef EXPLORE_MODE
```
DarkShinz: This update is an important improvement than just to be defined by the user. If you are uncertain if your code really works, allow me to give you a few tips.

src/client/minimap.cpp
```diff
@@ -203,12 +209,100 @@ void Minimap::updateTile(const Position& pos, const TilePtr& tile)
...
+	//First we need to get the player object
+	LocalPlayerPtr localPlayer = g_game.getLocalPlayer();
+
+	// Do nothing in case the player object is not valid
+	if (localPlayer == nullptr) {
+		return;
+	}
+
+	// Get the current Z-Position and cache the current Player offset based on the player position and tile position
+	short playerPosZ = localPlayer->getPosition().z;
```
DarkShinz: I recommend doing it this way: 
`const Position& playerPos = localPlayer->getPosition();`
So you can use x, y, z without calling it again. So whenever you need to do a check you just call 'playerPos.z'

src/client/minimap.cpp
```diff
@@ -203,12 +209,100 @@ void Minimap::updateTile(const Position& pos, const TilePtr& tile)
...
+
+			//Looks like we passed all checks and are capable to view this tile. Lets draw the minimap data
+			it->block->updateTile(it->pos.x - it->offsetPos.x, it->pos.y - it->offsetPos.y, it->minimapTile);
+			it->block->justSaw();
+			it->drawn = true;
+			m_minimapFloorCache[playerPosZ].erase(it);
+			it = cache.begin();
+		}
+	}
+#endif
```
DarkShinz: Another tip is to create functions for code that intend to do something different than what the function describes. For example, checking if the MinimapTile should be removed from the cache, should be a function named like `updateMinimapCache()`. This way ppl reading your code would understand what the function does without the need to read comments.

Regarding the code functionality, I want to add my opinion that checks into a cache should be done ONLY when needed. Checking if the cache should be edited every single call of the Minimap::draw function should make significant overhead and impact player's FPS. Take care with those kinds of code.
With that in mind, I believe that a function to check the `updateMinimapCache ` should be called whenever the player position changes or the g_map.tiles updates (taking in consideration you are using the players position to check if minimap tiles are in range. It would be a waste to keep checking it if the position hasn't changed at all. See?).

src/client/minimap.cpp
```diff
@@ -203,12 +209,100 @@ void Minimap::updateTile(const Position& pos, const TilePtr& tile)
...
+		//Create new cache object
+		MinimapCache minimapCache;
+		minimapCache.drawn = false;
+		minimapCache.minimapTile = minimapTile;
+		minimapCache.pos = pos;
+		minimapCache.block = &block;
+		minimapCache.offsetPos = offsetPos;
+
+		//In case the player object is valid and we are capable to view this tile, draw it without caching
+		if (playerPosZ == pos.z && std::abs(localPlayer->getPosition().x - pos.x) <= 9 && std::abs(localPlayer->getPosition().y - pos.y) <= 7) {
```
DarkShinz: I see two improvements here:

1.  You need to get the current View Range of the client to use later in position checks. OTC codes should be flexible to this kind of customization, it would look like this:
`AwareRange range = g_map.getAwareRange();`

2. You'd rather use Position::isInRange rather than this check.
It would be something like this (using the previous review about player pos and the previous item):
`if (playerPos.isInRange(pos, range.horizontal(), range.vertical(), 0))`

VagnerDomingues:
interesting

spekdrum:
Hi. I've tested this code and causes a freeze in some cases. For example if you go to a multilevel map zone and go to the top and then go down to the level 6.

LordHepipud:
Thank you for the post. I will have to re-check, but I believe I already run into a similar issue and fixed it locally.
As the feedback was quite low, I expected no one is using it :D

I will keep you updated

Klonoa88:
Same here. This code sometimes freezes when walking between floors below floor 7

Klonoa88:
> 
> 
> Thank you for the post. I will have to re-check, but I believe I already run into a similar issue and fixed it locally.
> As the feedback was quite low, I expected no one is using it :D
> 
> I will keep you updated

Same here. This code sometimes freezes when walking between floors below floor 7

LordHepipud:
I just made a commit with some changes. Mainly the issue was caused by

```c++
if (it->drawn == true) {
    m_minimapFloorCache[playerPosZ].erase(it);
    it = cache.begin();
    continue;
}
```

which is not handled differently. Can you please test if the new code works/compiles? I have an issue with the old version of my source and sadly cant compile it for testing if I missed anything.
This part is working on my server without any freezes and crashes for a very long time.

modules/corelib/ui/uiscrollbar.lua
```diff
@@ -28,7 +28,7 @@ local function calcValues(self)
     proportion = math.min(math.max(self.step, 1), range)/range
   end
 
-  local px = math.max(proportion * pxrange, 10)
+  local px = math.max(proportion * pxrange, 6)
```
slavidodo, line 31: is this relevant to the commit?

src/client/minimap.cpp
```diff
@@ -24,6 +24,9 @@
 #include "minimap.h"
 #include "tile.h"
 
+//PhiadariaSoft Mod: Minimap drawing fix to prevent drwaing multiple floors
```
slavidodo, line 27: remove that line or add it to where it fits

src/client/minimap.cpp
```diff
@@ -96,6 +99,9 @@ void Minimap::clean()
 {
     for(int i=0;i<=Otc::MAX_Z;++i)
         m_tileBlocks[i].clear();
+        
+    m_minimapFloorCache.clear();
```
slavidodo, line 103: indentation

src/client/minimap.cpp
```diff
@@ -203,12 +209,91 @@ void Minimap::updateTile(const Position& pos, const TilePtr& tile)
         minimapTile.speed = std::min<int>((int)std::ceil(tile->getGroundSpeed() / 10.0f), 255);
     }
 
-    if(minimapTile != MinimapTile()) {
+#ifndef EXPLORE_MODE
+    if (minimapTile != MinimapTile()) {
```
slavidodo, line 213: follow the style here to make it compatible with all the code!
``if(x)``

src/client/minimap.h
```diff
@@ -73,6 +73,15 @@ class MinimapBlock
     stdext::boolean<false> m_wasSeen;
 };
 
+//EXPLORATION MODE
```
slavidodo, line 76: use ```#ifdef EXPLORE_MODE``

src/client/minimap.h
```diff
@@ -107,6 +116,14 @@ class Minimap
                                                                   (index / (65536 / MMBLOCK_SIZE))*MMBLOCK_SIZE, z); }
     uint getBlockIndex(const Position& pos) { return ((pos.y / MMBLOCK_SIZE) * (65536 / MMBLOCK_SIZE)) + (pos.x / MMBLOCK_SIZE); }
     std::unordered_map<uint, MinimapBlock> m_tileBlocks[Otc::MAX_Z+1];
+    
+    //EXPLORATION MODE
```
slavidodo, line 120: use ```#ifdef EXPLORE_MODE``

src/client/minimap.cpp
```diff
@@ -203,12 +209,91 @@ void Minimap::updateTile(const Position& pos, const TilePtr& tile)
...
+            }
+
+            // Looks like we passed all checks and are capable to view this tile. Lets draw the minimap data
+            it->block->updateTile(it->pos.x - it->offsetPos.x, it->pos.y - it->offsetPos.y, it->minimapTile);
+            it->block->justSaw();
+            m_minimapFloorCache[playerPosZ].erase(it);
+            it = cache.begin();
+        }
+    }
+#endif
```
slavidodo, line 296: follow the style we're using in the code! single-line if statement, ...

modules/corelib/ui/uiscrollbar.lua
LordHepipud: No this shouldn't be a part of it. No idea how this was added Oo

src/client/minimap.cpp
LordHepipud: What? Oo
Where is the problem here? Oo

slavidodo:
```cpp
if(x)
    DoSomething();

for(x){
```

src/client/minimap.cpp
slavidodo: @LordHepipud just like here :)

diath:
A much simpler approach would be to simply not update the minimap when the z coordinate is different, I don't see much reason to cache the previously seen tiles, if the player wasn't to that place yet then it makes sense that they should discover that area on their own once their get to the destination floor:
```diff --git a/src/client/map.cpp b/src/client/map.cpp
index 3f768c257234b9dc79784f52460c170b34c46543..6f7db24229e9e9a54f5abaf14cf6860afd40d08c 100644
--- a/src/client/map.cpp
+++ b/src/client/map.cpp
@@ -44,7 +44,11 @@ void Map::notificateTileUpdate(const Position& pos)
 
     for(const MapViewPtr& mapView : m_mapViews)
         mapView->onTileUpdate(pos);
-    g_minimap.updateTile(pos, getTile(pos));
+
+    LocalPlayerPtr localPlayer = g_game.getLocalPlayer();
+    if (localPlayer && localPlayer->getPosition().z == pos.z) {
+        g_minimap.updateTile(pos, getTile(pos));
+    }
 }
 
 void Map::clean()
```

Also it would be better if this was behind a runtime flag, not a compile time flag.

Issue #599, opened 2014-11-19
Bug after die.


Anastaciaa:
Hello,
Today i noticed bug it happens on servers < 8.0 
It just look like you can walk 1 sqm after you die.

PR #601, opened 2014-11-20
Fixing issue with actions after player death

LordHepipud:
After a death of the player you still could move one tile away from your corpse making the client to draw the creature object again including all light effects. Up to that you still could perform actions with the mouse which the client displayed.
Up to that a cleanup was made, initialising the player object once he log ins preventing to have to call the function all over again.
Fixing issue #599

### DIFF
--- a/modules/game_interface/gameinterface.lua
+++ b/modules/game_interface/gameinterface.lua
@@ -281,11 +281,9 @@ end
 
 function changeWalkDir(dir, pop)
 
-  if player then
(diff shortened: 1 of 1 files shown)


BenDol:
This still an issue? Never experienced it myself.

iryont:
It does happen on servers in which server does not confirm the death (m_death variable). Like I said in my previous message - this specific fix was way too complicated.

PR #632, opened 2015-02-05
Update on textmessage.otui

SuggestName:
Change in the anchor statusLabel so that it is always on top of the console and never behind him.
When you press Ctrl + . you can not view the error messages .

Current version: http://prntscr.com/61k0wb
Correction: http://prntscr.com/61k0l3


edubart:
Combine all commits into one, then I will accept the request
**TIP:** You can use `git rebase -i HEAD~5` to combine 5 commits into one.

SuggestName:
Like this ? i make it right ?

PR #654, opened 2015-04-06
Market myoffers + NPC start block

crackcomm:
My offers in market are now shown and can be canceled. What is missing:
You have to refresh offers after You go to "My offers" to get them. No item can be selected when refreshing (use reset).

### DIFF
--- a/modules/game_market/marketprotocol.lua
+++ b/modules/game_market/marketprotocol.lua
@@ -29,11 +29,12 @@ local function readMarketOffer(msg, action, var)
   local state = MarketOfferState.Active
   if var == MarketRequest.MyHistory then
     state = msg:getU8()
+  elseif var == MarketRequest.MyOffers then
   else
     playerName = msg:getString()
   end
 
-  return MarketOffer.new({timestamp, counter}, action, Item.create(itemId), amount, price, playerName, state)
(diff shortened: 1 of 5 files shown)


modules/game_market/marketprotocol.lua
```diff
@@ -29,11 +29,12 @@ local function readMarketOffer(msg, action, var)
   local state = MarketOfferState.Active
   if var == MarketRequest.MyHistory then
     state = msg:getU8()
+  elseif var == MarketRequest.MyOffers then
```
BenDol: Can be removed since nothing is being done here

BenDol:
:+1:

modules/game_market/marketprotocol.lua
crackcomm: It should replace the else statement
```elseif var ~= MarketRequest.MyOffers then```
because If it's my offer we can't read the string.

iryont:
:+1: 

Just remember to name variables correctly, e.g. MarketOfferState.Active -> marketOfferState.active

crackcomm:
btw. is offer exhaust really required?

TheSumm:
I think it is currently used to prevent sending another request before the server responded.
So it would not be required if everything is implemented correctly.

PR #665, opened 2015-05-25
Use native optimizations instead of hardcoded defaults

ranisalt:
Just like I noted in [this pull request](https://github.com/otland/forgottenserver/pull/1310) on otland/forgottenserver, we should use `-march=native` to turn on aggressive optimizations on best-performance builds. There were some hardcoded defaults in the CMakeLists which I have replaced with the improved flag.


iryont:
No, we should not really use it. The client is redistributable and should work on all CPUs (including those without AVX/SSE), while the server is meant to be as efficient as possible on the target machine.

ranisalt:
What I understand is when you compile for performance, you are trying to get the best crafted executable for your machine only. There is another target, RELEASE, that compiles with good generic `-O2` flag if anyone wants to release the executable to other people.

PR #671, opened 2015-07-05
Fix #664

### LINKED ISSUE #664: why usable flag change elevation?
otclient/src/client/thingtype.cpp line 247 "case ThingAttrUsable:"
I see the rme editor code, the usable flag does not modify elevation. Is this a bug?

Mignari:


iryont:
Again, I'm not familiar with newest protocol, however, since this flag was apparently meant to indicate 2 extra bytes of data (what kind of?) it will cause problems. It would be wise for someone to explain for what this flag is used in new Tibia and how to proceed with apparently two extra bytes of data.

Mignari:
It still read 2 bytes, but not changes the elevation.

TheSumm:
@iryont The property is used for Flash client to determine whether to use "Look" or "Use" when using the smart-click option.
It seems like the line was just added in the wrong place here https://github.com/ottools/otclient/commit/c81a623c43278af528adbd1dac0bf6b3cd7eeb38 by mistake.

TheSumm:
Thanks @Mignari

iryont:
My bad, I did not notice it. Good job.

PR #676, opened 2015-07-17
Serverlist Fix

Quintinon:
There was an issue where I had a server saved in the ServerList with no port. When this occurred it prevented the serverlist module from loading compeltely. I've fixed it such that if no port or protocol can be retrieved the server it was saved for is assumed bad and removed. Loading continues afterwards.


TheSumm:
Thanks for raising the issue, I fixed it here https://github.com/edubart/otclient/commit/b822e92c0e4f8935c39c094900492877bd2dfd1b as I did not like how ServerList.load was done.

PR #677, opened 2015-07-17
Change when load and save occurs in gameinterface

Quintinon:
Change when gameinterface calls load and save to prevent bottomSplitter from being put in the wrong location due to initial load resize events.

What was occurring was:
gameinterface module loaded and set bottomSplitter (between game window and chat) to the saved position.
onGeometryChanged event was called 3 times (I could not find the root cause of startup calling this, but probably due to loading if the window should be fullscreen or not. My client was setup to load to fullscreen, while windowed mode was very small). The onGemoetryChanged event changed where the bottomSplitter was (probably due to the small windowed mode that the game may "open" with for a few milliseconds).

Now what happens is:
gameinterface module loads
initial onGeometryChanged events are called
when g_app.onRun is called the splitter settings are loaded.

I moved the save to occur on the g_app.onExit for symmetry.

### DIFF
--- a/modules/game_interface/gameinterface.lua
+++ b/modules/game_interface/gameinterface.lua
@@ -26,7 +26,12 @@ function init()
     onGameEnd = onGameEnd,
     onLoginAdvice = onLoginAdvice,
   }, true)
-
+  -- Call load AFTER game window has been created and resized to a stable state, otherwise the saved settings can get overridden by false onGeometryChange events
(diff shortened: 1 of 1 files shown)


BenDol:
Have you had any load issues doing it this way? Race conditions etc.

Quintinon:
I haven't noticed any issues loading this way

PR #678, opened 2015-07-24
Replace deprecated function

ranisalt:
`RSA* RSA_generate_key` is deprecated from OpenSSL [(source)](https://www.openssl.org/docs/crypto/RSA_generate_key_ex.html). I replaced it with the modern version.

Without this change, it does not compile on bleeding edge Linux distros such as Arch Linux.


iryont:
Good job.

PR #680, opened 2015-08-26
Add a flag for idle animation game feature.

Mignari:
This is necessary to disable the feature when necessary.

https://otland.net/threads/10-76-animation-bug.233461/


iryont:
It does look fine to me, even though I'm not really familiar with newest protocols.

If you are up to finishing your work then I will gladly merge the rest as well.

PR #683, opened 2015-08-31
remake Visual Project

Radseq:
changes in market: some windows are larger
buttons etc, will resize depending on client language

iam using boost from http://sourceforge.net/projects/boost/files/boost-binaries/1.57.0/
boost_1_57_0-msvc-12.0-32.exe

this changes in vc12 make possibility to debug, debug working on my own compiled dll like libvorbis
sdk allowed by you dont work to do debug, for release is good

So loords of otclient :smile_cat:  please relase new sdk for otclient (32bit with debug info and without, and 64bit with debug info and without)
and be free to using breakpoint.

new Edit!!!:
you dont have to do new sdk i missed one dll
be sure you check: tools->options->debugging->symbols 
now you can debug !!!!

### DIFF
--- a/modules/game_market/market.otui
+++ b/modules/game_market/market.otui
@@ -1,7 +1,7 @@
 MarketWindow < MainWindow
   id: marketWindow
   !text: tr('Market')
-  size: 700 530
+  size: 800 530
 
   @onEscape: Market.close()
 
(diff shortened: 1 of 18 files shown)


BenDol:
LGTM

PR #690, opened 2015-10-06
Add 'GameKeepUnawareTiles' feature.

Mignari:
Remove tiles that we are not aware anymore by default.

### DIFF
--- a/modules/gamelib/const.lua
+++ b/modules/gamelib/const.lua
@@ -132,6 +132,7 @@ GameUnjustifiedPoints = 68
 GameSessionKey = 69
 GameDeathType = 70
 GameIdleAnimations = 71
+GameKeepUnawareTiles = 72
 
 TextColors = {
   red       = '#f55e5e', --'#c83200'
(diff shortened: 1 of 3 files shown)


TheSumm:
Great! Thank you.

iryont:
The idea is good, but it's highly inefficient.

hrsantiago:
You wont have a lot of tiles, cause it's cleaned every time the map position changes. The maximum amount of tiles is around 18 * 15 * 16 = 4320. However this will be really uncommon. Usually, floor 7 will be completely full and maybe 2 more floors with some tiles. I'd say ~500 tiles is a good average, which is not a huge loop to do for each step.

iryont:
Unfortunately you are assuming that everyone uses a view of 18x14 tiles. In such case it will work just fine, but it's still an inefficient code for larger views.

hrsantiago:
Yeah, but i guess this is the default so most people wont have issues with it. For those using larger views, they've changed the map aware range somewhere. At this same place they can simply disable this feature.

PR #702, opened 2016-01-08
Log file saves specific to server and character

Sportacus:
Changed it so when you save a log file, the log file name includes the
server name and the characters name, in the file name.

This is to improve organization, to separate logs based off the server
you are on, and the character you are logged in.


edubart:
Merged in https://github.com/edubart/otclient/commit/a9f6f7662f21cedf1b171545dc1bf2de1188abcc

PR #704, opened 2016-01-11
Fix annoying warnings

Nu77:
Not tested in Linux or Windows, but should work fine.


hrsantiago:
Which compiler?

Nu77:
```code
Apple LLVM version 7.0.2 (clang-700.1.81)
Target: x86_64-apple-darwin15.2.0
Thread model: posix
```

PR #706, opened 2016-01-11
Turn off XML

Nu77:
Why is XML turned on by default since it's used in framework and not on client?


Nu77:
Well, now i know why it's turned on. XML it's not defined, but always required. Don't know why though.

hrsantiago:
It's included for mapeditor stuff. @asamy45 was working on a mapeditor using otclient as a framework.

PR #718, opened 2016-02-18
Remake Visual Project

Radseq:
No tested i just give you what i do last time:
https://github.com/edubart/otclient/pull/683
I close becouse no one wants marge
last time was working, so i hope is working now


Radseq:
So, i dont know maybe some day i will back to otclient
If you want you can close, but i strongly recommend to marge becouse is organize and transparent

PR #748, opened 2016-07-26
Fixes #743

### LINKED ISSUE #743: NPC-Chat wont show up.
Hello all,

using newest release of otclient on 4.6.2-1-ARCH, self compiled, i noticed that i cant talk with any npc in game. Its seems like they still react to the "hello" or "hi" because they turn in my direction.
but the npc chat wont open and no trade windows or further interaction is possible, like depositing all my gold.

The following error appears in the otclient logs:
`ERROR: ProtocolGame parse message exception (132 bytes unread, last opcode is 170, prev opcode is 170): InputMessage eof reached
ERROR: ProtocolGame parse message exception (14 bytes unread, last opcode is 170, prev opcode is 109): InputMessage eof reached
ERROR: ProtocolGame parse message exception (207 bytes unread, last opcode is 170, prev opcode is 170): InputMessage eof reached
`

The value befor "bytes" varies as you can see, but "last opcode" ist always 170, also "prev opcode".

I hope someone knows a solution for this because using otclient seems like the only real way to play on ot-tibia servers on linux. and not being able to talk to npc's sucks.

Iam playing on a server with the game version 8.6, but also on higher versions this bug appears.

greetings

eklundkristoffer:


conde2:
@assamy, @TheSumm

This can be closed

PR #750, opened 2016-08-15
Some possible Fixes when compiling on Windows

Marcelo-Petrucelli:
Some possible fixes when compiling on Windows, such as using **WIN32** and using -U__STRICT_ANSI__ because of Boost on MinGW.

I had other problems dealing with MinGW and Codeblocks, but those solutions solved some major compiling problems.


edubart:
This is hopefully fixed in https://github.com/edubart/otclient/commit/8309fee510a30cb335813d246080ae3a0858a889

PR #771, opened 2016-11-08
Fix inconsistent coding style

jo3bingham:


iryont:
That is lovely :) Thanks. It is always good to keep the coding style consistent.

PR #778, opened 2016-11-15
Update const.lua

A-Syntax:
Players were not able to speak to the npc.
https://github.com/edubart/otclient/issues/743#issuecomment-236684402


conde2:
@TheSumm, @assamy

This can be closed

PR #802, opened 2017-01-13
Update Visual Studio build files

brunominervino:


conde2:
Thats really great, have you managed to compile the libs to MSVC 2015?

brunominervino:
@asamy and @TheSumm, anyone can confirm this changes? I have made this changes using boost_1_60_0, but I got error with libvorbis v1.3.3. Now, I'm at work and don't have more time to check the compatibility. Any of you can confirm this? Thank you.

conde2:
I can confirm that the libs need to be updated, you will need to build the libs to work on MSVC14

asamy:
The libraries definitely need to be updated, but why you can probably build them easily?  Maybe use TFS SDK?

brunominervino:
Maybe this is a good idea @asamy. Soon as possible I'll try, but, if anyone want to try first, please give us a feedback.

conde2:
I don't think it is anyway easy to build the libs since most of them don't have MSVC projects

brunominervino:
@conde2, some labs have newest versions, like vorbis:
http://downloads.xiph.org/releases/vorbis/

conde2:
Yes we definitely should update the libs, I'm just saying that it will be really hard to update all of them, because most of them don't have visual studio projects that we can simply compile from them

brunominervino:
Definitely, you're right. Maybe @edubart can give to us a start.

conde2:
LuaJit we can use the same from TFS
Boost we can download it from boost website
Zlib does have a MSVC project to build it
Glew I think it can be downloaded from website not sure

the problem is

Vorbis / Ogg
OpenSSL
physfs
OpenAL

conde2:
So far i managed to compile this:

image

brunominervino:
@conde2 great!

conde2:
Here are some stuff on how to compile OpenAL and Physys

http://supertux.lethargik.org/wiki/Building_on_Windows

conde2:
Ok just some more and we are done =)

image

brunominervino:
Congrats man!!!

conde2:
image


DONE !
I will just try the x64 version and update the project and libs, It will also be good if someone update the wiki with the new libs and how to install them

conde2:
I had some troubles with Zlib, I will try fix this by monday, and then when everything is working I will pull a commit.

Please don't revert this commit yet

brunominervino:
Nice work @conde2, thank you!

PR #817, opened 2017-02-01
Update README

conde2:
Remove all broken links and fix MSVC link


conde2:
@TheSumm 

Could you build the latest OTClient and update the OTLand thread ?

PR #818, opened 2017-02-01
Fix Look

conde2:
sendLookCreature should only be called in battleList, checked with TFS 10.98 and downgraded versions


Steini64:
Yesterday I wondered why I receive the packet in onPlayerLookBattleWindow() when looking on creatures and was puzzled for half an hour why my changed lua-code in server had no effect until I figured out it arrives via battle-window look. I will try this fix. Thanks so far!

PR #844, opened 2017-05-12
Fix iptostring in LUA

gesior:
Fix iptostring bug that blocks access to half of OTSes


src/framework/luafunctions.cpp
```diff
@@ -72,7 +72,7 @@ void Application::registerLuaFunctions()
     g_lua.bindGlobalFunction("pointtostring", [](const Point& v) { return stdext::to_string(v); });
     g_lua.bindGlobalFunction("colortostring", [](const Color& v) { return stdext::to_string(v); });
     g_lua.bindGlobalFunction("sizetostring", [](const Size& v) { return stdext::to_string(v); });
-    g_lua.bindGlobalFunction("iptostring", [](int v) { return stdext::ip_to_string(v); });
+    g_lua.bindGlobalFunction("iptostring", [](uint v) { return stdext::ip_to_string(v); });
```
iryont: Make it uint32 instead and I will merge it.

gesior:
Changed to uint32

src/framework/luafunctions.cpp
gesior: Done

iryont:
Good job.

PR #861, opened 2017-07-27
Add proper FindOpenSSL version, force version 1.0

ranisalt:
The current code of `FindOpenSSL.cmake` is improper and does not follow user definitions, and this is especially awful with the (shamelessly minor) update of OpenSSL to 1.1. It means we can not set a path for a custom located OpenSSL 1.0 build, rather relying only on the system-wide one, where 1.1 is already spreading.

I fixed it replacing with packaged `FindOpenSSL.cmake` from CMake 3.8, please shout if it does not work will older versions.

To compile with OpenSSL version, have it installed elsewhere (on Arch, I have the package `openssl-1.0` which installs to `/usr/lib/openssl-1.0`) and run the following configuration line:

```
$ cmake -DOPENSSL_INCLUDE_DIR=/usr/include/openssl-1.0 -DOPENSSL_LIBRARIES=/usr/lib/openssl-1.0 ...
```

Of course, adapting it to your install. `OPENSSL_INCLUDE_DIR` is where the headers go, it should contain a directory named `openssl` with the headers inside. `OPENSSL_LIBRARIES` are where the library files (`.so`) are.

### DIFF
--- a/src/framework/CMakeLists.txt
+++ b/src/framework/CMakeLists.txt
@@ -219,14 +219,12 @@ endif()
 message(STATUS "LuaJIT: " ${LUAJIT})
 
 find_package(PhysFS REQUIRED)
-find_package(OpenSSL REQUIRED)
 find_package(ZLIB REQUIRED)
 
 set(framework_LIBRARIES ${framework_LIBRARIES}
     ${Boost_LIBRARIES}
     ${LUA_LIBRARY}
     ${PHYSFS_LIBRARY}
-    ${OPENSSL_LIBRARIES}
     ${ZLIB_LIBRARY}
 )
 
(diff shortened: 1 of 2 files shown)


ranisalt:
It will not build on Travis until machine is upgraded to Trusty. Old CMake has erratic behavior and does not consider `1.0.1` to match exact version `1.0`, while CMake 3.0+ does it properly, e.g. on Arch Linux with OpenSSL 1.0.2l from `openssl-1.0` package:

```
-- Found OpenSSL: /usr/lib/openssl-1.0/libcrypto.so (found suitable exact version "1.0.2l") 
```

EvulMastah:
Why not just update the code to work with newer OpenSSL?
Here is the example code:
```cpp
void Crypt::rsaSetPublicKey(const std::string& n, const std::string& e)
{
#if OPENSSL_VERSION_NUMBER < 0x10100000L
	BN_dec2bn(&m_rsa->n, n.c_str());
	BN_dec2bn(&m_rsa->e, e.c_str());

	// clear rsa cache
	if (m_rsa->_method_mod_n) {
		BN_MONT_CTX_free(m_rsa->_method_mod_n);
		m_rsa->_method_mod_n = nullptr;
	}
#else
	BIGNUM *rsa_n = BN_new();
	BIGNUM *rsa_e = BN_new();

	BN_dec2bn(&rsa_n, n.c_str());
	BN_dec2bn(&rsa_e, e.c_str());

	RSA_set0_key(m_rsa, rsa_n, rsa_e, nullptr);
#endif
}
```

ranisalt:
I tried, but got a lot of errors. OpenSSL documentation is pure shit and I don't have the time to update it myself.

edubart:
In https://github.com/edubart/otclient/commit/8b161e1a07a16f4c6af49575a9a92a8a103ec029 this is done (try to find OpenSSL 1.0 if available)

PR #865, opened 2017-08-28
Fixed items minimap color

Erexo:
Some items marked as "common" are actually an important terrain items (like doors/fences/street lamps and so on). It makes minimap much more readable if every item with minimap color is shown (as in Cip client).


Erexo:
Known bug;
Since there is one opcode from server that tells to stop both attacking and following, when player is following a creature and decide to attack other, the client will remove attacked crature and add followed, trigger event that we are not attacking anything now so server send us back this cancel opcode so client cancel the follow and we end up with nothing. Cip client somehow managed to pass through it

Felipe1993:
so this passed or not?

andersonfaaria:
Instead of deleting the if statement you should have checked for the specified items. If this is merged we will have to iterate through every item to draw it on map, which would be very bad. Also you should make different branches for different commits instead of just make one pull request with lots of commits. As for the bug of attacking who you are following, I'll test it to confirm

Felipe21:
Any updates?

Issue #697, opened 2015-12-25
Shooting runes in battlelist @ creatures.


Warchas:
I am using protocol 10.41 and I have disabled hotkeys on my server.
When I am trying to aim runes at for example giant spiders that move very fast it will often say "You can only use this rune at creatures."

I seems like the client is trying to shoot the rune at the SQM that the monster was at right when I press but the same second it has moved and therefore it "misses" and generates the error message "You can only use this rune at creatures."

Is there some way to fix this?

Sincerely, 
Warchas

PR #874, opened 2017-10-23
Fixed problem with function "useWith" on creatures

EgzoT:
Fixed problems with selecting moving monsters in battle list.
Fixed problems with using hotkeys item on creatures (protocol < 780).

Answer to issues:
[#856](https://github.com/edubart/otclient/issues/856)
[#697](https://github.com/edubart/otclient/issues/697)

Threads with this problem on OTLand:
[Problem with Battle - Monsters.](https://otland.net/threads/problem-with-battle-monsters.247449/)
[Can only use this rune on creatures [Bug?]](https://otland.net/threads/can-only-use-this-rune-on-creatures-bug.239093/)

### DIFF
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -856,7 +856,10 @@ void Game::useWith(const ItemPtr& item, const ThingPtr& toThing)
     if(!pos.isValid()) // virtual item
         pos = Position(0xFFFF, 0, 0); // means that is an item in inventory
 
-    m_protocolGame->sendUseItemWith(pos, item->getId(), item->getStackPos(), toThing->getPosition(), toThing->getId(), toThing->getStackPos());
+    if(toThing->isCreature())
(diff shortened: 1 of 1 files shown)


hrsantiago:
How do you know that it fixes those issues if you didn't test it?

EgzoT:
I compared action of functions "useWith" and "useInventoryItemWith", and function "useInventoryItemWith" work perfectly with creatures (look lines 869-872). But this function is a little other (don't send full item position like useWith, sends code 0, I think this is code to finding item in inventory on server side).
After compare I have seen that function useWith is deficient, function useWith does not distinguish use on creature and use on item.

EgzoT:
Compiled and tested, all works fine.
Now battle list/hotkeys(in protocol < 780) using on creatures works perfectly like in original client.

edubart:
Looks fine.

PR #875, opened 2017-10-23
*Fix Talking with NPC

AngelOfDeaths:
i don't know why someone change it but its wrong, now will be correct.


hrsantiago:
Most likely this one should be outside the switch, with an if to check the protocol version

edubart:
This fix looks wrong, I think you fixed just for the protocol you are using protocol, this may depend on the server protocol version.

PR #878, opened 2017-10-25
[Probable bug] Double use function onAppear

EgzoT:
Method Map::addThing uses two times function onAppear for object Missile.
First used in line 119, but then used again the same function in line 165.
Without first function onAppear method Map::addThing work the same.


edubart:
Good catch.

Issue #837, opened 2017-04-28
OpenSSL API Change


diath:
### Steps to reproduce (include any configuration/script required to reproduce)
1. Build otclient against OpenSSL 1.1.0x version

### Expected behaviour
The crypto module does compile.

### Actual behaviour
The crypto module does not compile.

### Detail
Basically what happens is that the OpenSSL 1.1.0x and up changed the API and broke backwards compatibility. The RSA struct members are no longer accessible directly. Currently one can just build otclient against the older version of OpenSSL but eventually distros will start shipping the newer version so I'm posting it as something to keep in mind.

Some more information on the topic:
https://wiki.openssl.org/index.php/OpenSSL_1.1.0_Changes
http://stackoverflow.com/a/40550964

ranisalt:
The hard part is OpenSSL documentation is completely useless. It literally does not document anything other than public signature of methods, which anyone already sees.

I think implementing RSA without OpenSSL, on a fashion similar to forgottenserver, is the way to go. OpenSSL code is complete spaghetti.

gesior:
I would like to see Crypt.cpp without OpenSSL. Can't compile it on Linux Mint (OpenSSL 1.1.0f).

If you know how to make it work without OpenSSL or with OpenSSL 1.1.0f and you need 15$:
https://www.bountysource.com/issues/44639063-openssl-api-change [15$ bounty]

-----
Add 'Bounty' label, like on forgottenserver repo. I'm sure that there are many people ready to pay for bug fixes and new features in otclient.

diath:
For the time being you can simply install OpenSSL 1.0x and point CMake to that OpenSSL installation directory instead but yeah ideally we'd either get rid of OpenSSL as a dependency or make it work with both API versions.

edubart:
In https://github.com/edubart/otclient/commit/8b161e1a07a16f4c6af49575a9a92a8a103ec029 I've made to prefer OpenSSL 1.0 over 1.1

JSkalskiSBG:
@diath  @edubart 
I downloaded, compiled and installed OpenSSL 1.0.2:
	
	q@JSkalski /usr/lib/x86_64-linux-gnu $ ls -lah | grep libcry
	-rw-r--r--   1 root root   60K Jun 16 22:58 libcrypt.a
	-rw-r--r--   1 root root  4.7M Jul  6 13:58 libcrypto.a
[...]

PR #888, opened 2017-11-09
WIP: Replace OpenSSL with GMP for RSA

ranisalt:
That should fix recent complaints about OpenSSL linkage/compilation issues #837 #861. It should be tested more thorougly, but it should work simetrically with TFS.

I also removed functions that were not being used and depended on OpenSSL.

### DIFF
--- a/src/framework/util/crypt.h
+++ b/src/framework/util/crypt.h
@@ -28,7 +28,7 @@
 
 #include <boost/uuid/uuid.hpp>
 
-typedef struct rsa_st RSA;
+#include <gmp.h>
 
 class Crypt
 {
@@ -44,17 +44,14 @@ public:
     std::string genUUID();
     bool setMachineUUID(std::string uuidstr);
     std::string getMachineUUID();
-    std::string md5Encode(const std::string& decoded_string, bool upperCase);
-    std::string sha1Encode(const std::string& decoded_string, bool upperCase);
-    std::string sha256Encode(const std::string& decoded_string, bool upperCase);
-    std::string sha512Encode(const std::string& decoded_string, bool upperCase);
 
-    void rsaGenerateKey(int bits, int e);
+    static constexpr auto MODULUS_SIZE = 1024;
+    static constexpr auto BLOCK_SIZE = MODULUS_SIZE / 8;
+
     void rsaSetPublicKey(const std::string& n, const std::string& e);
     void rsaSetPrivateKey(const std::string &p, const std::string &q, const std::string &d);
-    bool rsaCheckKey();
-    bool rsaEncrypt(unsigned char *msg, int size);
-    bool rsaDecrypt(unsigned char *msg, int size);
+    bool rsaEncrypt(char *msg, int size);
+    bool rsaDecrypt(char *msg, int size);
     int rsaGetSize();
 
 private:
@@ -62,7 +59,7 @@ private:
     std::string _decrypt(const std::string& encrypted_string, bool useMachineUUID);
     std::string getCryptKey(bool useMachineUUID);
     boost::uuids::uuid m_machineUUID;
-    RSA *m_rsa;
+    mpz_t n, e, d;
 };
 
 extern Crypt g_crypt;
--- a/src/framework/util/crypt.h
+++ b/src/framework/util/crypt.h
@@ -27,8 +27,11 @@
 #include <string>
 
 #include <boost/uuid/uuid.hpp>
-
+#ifdef USE_GMP
 #include <gmp.h>
+#else
+typedef struct rsa_st RSA;
+#endif
 
 class Crypt
 {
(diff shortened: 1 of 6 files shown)


src/framework/util/crypt.h
```diff
@@ -44,25 +44,22 @@ class Crypt
...
     bool setMachineUUID(std::string uuidstr);
     std::string getMachineUUID();
-    std::string md5Encode(const std::string& decoded_string, bool upperCase);
-    std::string sha1Encode(const std::string& decoded_string, bool upperCase);
-    std::string sha256Encode(const std::string& decoded_string, bool upperCase);
-    std::string sha512Encode(const std::string& decoded_string, bool upperCase);
 
-    void rsaGenerateKey(int bits, int e);
+    static constexpr auto MODULUS_SIZE = 1024;
+    static constexpr auto BLOCK_SIZE = MODULUS_SIZE / 8;
```
ranisalt: I could use `CHAR_BIT` instead for ｃｏｍｐａｔｉｂｉｌｉｔｙ, but would anyone ever use OTClient on a DSP?

edubart:
I don't see this as an improvement, you just replaced one dependency (OpenSSL) with another (GMP) while removing other code. While this could make easy to you compile, would require existing users to compile and install new dependency GMP.

If the problem was wanting OpenSSL 1.1 a better choice is to support it and not to introduce a new  library.

And OpenSSL always was better suited for this task.

ranisalt:
GMP is a much more consistent library (as you can see, OpenSSL introduces API breaking changes in MINOR releases, it's not even OpenSSL 2), is more performant and IMO has a much cleaner interface (though interfacing with C is painful anyway).

The problem is with every OpenSSL minor update you will have rewrite stuff and keep them for years to come, as happened with 0.x to 1.0 and now with 1.1. Some distros lag WAY behing others.

It of course can boil down to preference, but indeed I do prefer to depend on GMP for this very single task of RSA encryption than using full-fledged incompatibility-ridden OpenSSL.

gesior:
GMP functions names say what they do. There is almost no chance to make changes that break compability - these functions do simple math operations.

OpenSSL functions realize advanced algorithms that use many parameters. Someday they can decide that 'RSA without padding' (like in tibia) is insecure and it should not be available, as they did with SSL 2.0.

GMP functions used in RSA implementation ('mul' and 'powm') got same parameters and logic in all GMP versions from year 2000. You can use GMP 3.0 (from 2000) and this RSA implementation will work same as on 6.1 (2016). It does not matter what version you get with your linux distro.

HeavenIsLost:
I agree with gesior and ranisalt arguments. GMP is way more reliable when it comes about maintenance.

gesior:
Downloaded, compiled, works without OpenSSL. Ubuntu still with problems - GMP installed from repository does not work, had to compile newest version.

ranisalt:
@gesior what did not work and what versions did you test?

JSkalskiSBG:
I meant 'physfs', not 'gmp', little mistake. Default GMP is ok (libgmp3 [GMP 5.1.3]- installed with forgottenserver libs).

    /usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libphysfs.a: error adding symbols: Archive has no index; run ranlib to add one

PhysFS installed by default is version 2.0.3-3 (newest Linux Mint), somehow there are no symbols.
I ran 'ranlib' on some physfs .a file:

    ranlib /usr/lib/x86_64-linux-gnu/libphysfs.a

and then on linking stage it showed:

    CMakeFiles/otclient.dir/src/framework/core/filestream.cpp.o: In function `FileStream::size()':
    /home/q/Desktop/otclient-gmp/src/framework/core/filestream.cpp:162: undefined reference to `PHYSFS_fileLength'
    CMakeFiles/otclient.dir/src/framework/core/filestream.cpp.o: In function `FileStream::tell()':
    /home/q/Desktop/otclient-gmp/src/framework/core/filestream.cpp:170: undefined reference to `PHYSFS_tell'

I had to download PhysFS 2.0.3 source from website, compile and install.

-------------------------
I'm sure that this change is very good for Otclient. OpenSSL is a problem for many linux distros now and will be in the future.

ranisalt:
@JSkalskiSBG I am on physfs 3.0, I have no .a (static lib) file and it works ok. It is not related to this PR though.

Exedenes:
I just installed the GMP replacement on Kali linux because the OpenSSL was not installing properly.
But now i keep getting this error 2 ""Your connection has been lost.
Either your network or the server went down." whenever i try to log in to any servers. Its weird because it will also give other errors like "only clients with x protocol allowed" so it seems to be connecting to the servers at some level. Any idea why this might be?

JSkalskiSBG:
@Exedenes 
There is mistake in RSA implementation. It loses 1 character every encrypt/decrypt.
In file **crypt.cpp** in function **mpz_export** (in 2 places) there is **- count** change it to **- count - 1**
or try my pull request:
#918

Exedenes:
changing count doesn't seem to work, I'll try the pull request

Exedenes:
so i recompiled with otclient-gmp_or_openssl and im still having the same issue.

Exedenes:
NVM it works now at least with 8.6 servers. 7.4 and 7.72 seem to be broken still.

iryont:
Since we have compatibility for OpenSSL < 1.1.0 and >= 1.1.0 now I believe this PR is no longer needed. If anyone disagrees let me know, so we can discuss it.

ranisalt:
Indeed, but the GMP solution is much shorter and this PR also cleans some unused code.

iryont:
Agreed. I like the PR itself, but I would rather have a choice to either choose OpenSSL or GMP for that matter. Your code is much more clean than the previous one with support for both, so personally I would go with yours for the idea of having to choose between OpenSSL or GMP.

I will open this PR for now hoping someone wants to take care of it.

gesior:
@iryont 
> Agreed. I like the PR itself, but I would rather have a choice to either choose OpenSSL or GMP for that matter. Your code is much more clean than the previous one with support for both, so personally I would go with yours for the idea of having to choose between OpenSSL or GMP.
> 
> I will open this PR for now hoping someone wants to take care of it.

I wrote version with OpenSSL and GMP. It's a lot of code, but it was discusses in some issue, how this code should look like:
**crypt.cpp** includes one of classes from file **crypt_gmp.cpp** or **crypt_openssl.cpp**. I can add there some extra abstract class (interface) that will be extended (implemented) by both classes, but it would be more like programming art, not programming.

It could be also done in one crypt.cpp file with many **#ifdef** in each function, but it would look crappy.

Someone experienced in Makefiles could add there fallback option, if OpenSSL is not found.
https://github.com/edubart/otclient/pull/918/files

iryont:
I know, I saw that PR as well. However, having a look at PR of @ranisalt I can see there aren't many places with OpenSSL after removal of unused code, so having #ifdef wouldn't be that bad and would solve a lot of extra stuff.

Also, regarding fallback, something like this should work:

```find_package(OpenSSL QUIET)
if(NOT OPENSSL_FOUND)
    find_package(GMP REQUIRED)
    set(framework_DEFINITIONS ${framework_DEFINITIONS} -DUSE_GMP)
endif()```

I will check it out later on when I will have some free time.

iryont:
I did make some changes, but I haven't tested it. Just the idea of using OpenSSL & GMP with fallback.

iryont:
#1020

Issue #895, opened 2017-11-13
Messed up mapmark icons


andersonfaaria:
At least for newest versions of TFS, the mapmarks are all wrong
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/otclient.494/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Insert mapmarks by script into your minimap
2. Realize they are with wrong id

### Expected behaviour
        registerEnum(MAPMARK_TICK) = 0
	registerEnum(MAPMARK_QUESTION) =1
	registerEnum(MAPMARK_EXCLAMATION) = 2
	registerEnum(MAPMARK_STAR) = 3
	registerEnum(MAPMARK_CROSS) = 4
	registerEnum(MAPMARK_TEMPLE) = 5
	registerEnum(MAPMARK_KISS) = 6
	registerEnum(MAPMARK_SHOVEL) = 10
	registerEnum(MAPMARK_SWORD) = 11
	registerEnum(MAPMARK_FLAG) = 12
	registerEnum(MAPMARK_LOCK) = 13
	registerEnum(MAPMARK_BAG) = 14
	registerEnum(MAPMARK_SKULL) = 15
	registerEnum(MAPMARK_DOLLAR) =  16
	registerEnum(MAPMARK_REDNORTH) = 7
	registerEnum(MAPMARK_REDSOUTH) = 17
	registerEnum(MAPMARK_REDEAST) = 18
	registerEnum(MAPMARK_REDWEST) = 8
	registerEnum(MAPMARK_GREENNORTH) = 9
	registerEnum(MAPMARK_GREENSOUTH) = 19
### Actual behaviour
https://github.com/edubart/otclient/tree/master/data/images/game/minimap
Nothing like this

### Environment

Using latest rev of otc along with 10.76 version of tfs

iryont:
Can this issue be replicated by anyone else?

PR #896, opened 2017-11-13
Fix of #895 and variable name

andersonfaaria:
This is just a small renaming to images of minimap (the map marks) in order to sync the mark with the proper id from the tfs server. Otherwise, if you add marks by script, they will be wrong.
Also I've changed the variable name for oufit to outfit properly written.


EgzoT:
Probably you have right (flags are mixed up), tested on 8.6 OTS.
But you need to add description to this pull request, because if someone will see it, probably they will not know what you mean.
Add reference to issue, where you describe this problem. It will help to see your pull request with fix from this issue.

andersonfaaria:
Updated, I'm not so good with git since my team at work has always used visual studio for versioning. Hope it's ok now

edubart:
You pushed commits non ralated to the issue

PR #898, opened 2017-11-15
Add setPathable() function

EgzoT:
Adding function who changing item attribute (ThingAttrNotPathable).
Allows to walk on not pathable items (ThingAttrNotPathable -> true) like parcels, fire fields etc. and vice versa.
Modifying item attribute without editing items sources files. Temporary action for the duration of the client session.

Usage:
```
print(g_things.getThingType(3504):isNotPathable()) -> true

g_things.getThingType(3504):setPathable(true)

print(g_things.getThingType(3504):isNotPathable()) -> false
```

Helpful for highter levels players to ignoring "avoiding" fire/energy/poison fields etc. when using map click.

### DIFF
--- a/src/client/thingtype.h
+++ b/src/client/thingtype.h
@@ -178,7 +178,7 @@ public:
     bool isNotMoveable() { return m_attribs.has(ThingAttrNotMoveable); }
     bool blockProjectile() { return m_attribs.has(ThingAttrBlockProjectile); }
     bool isNotPathable() { return m_attribs.has(ThingAttrNotPathable); }
-	void setPathable(bool var) { var == true ? m_attribs.remove(ThingAttrNotPathable) : m_attribs.set(ThingAttrNotPathable, true); }
+	void setPathable(bool var);
     bool isPickupable() { return m_attribs.has(ThingAttrPickupable); }
     bool isHangable() { return m_attribs.has(ThingAttrHangable); }
     bool isHookSouth() { return m_attribs.has(ThingAttrHookSouth); }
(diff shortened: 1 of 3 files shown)


src/client/thingtype.h
```diff
@@ -178,6 +178,7 @@ class ThingType : public LuaObject
     bool isNotMoveable() { return m_attribs.has(ThingAttrNotMoveable); }
     bool blockProjectile() { return m_attribs.has(ThingAttrBlockProjectile); }
     bool isNotPathable() { return m_attribs.has(ThingAttrNotPathable); }
+	void setPathable(bool var) { var == true ? m_attribs.remove(ThingAttrNotPathable) : m_attribs.set(ThingAttrNotPathable, true); }
```
diath: I'd probably move the definition to the source file since it involves a conditional rather than being a simple proxy call. Also remove the extra indent.

EgzoT: Ok, I has moved definition from thingtype.h to thingtype.cpp file.
Previously i has inspired by function in line 158.

src/client/thingtype.h
```diff
@@ -178,6 +178,7 @@ class ThingType : public LuaObject
     bool isNotMoveable() { return m_attribs.has(ThingAttrNotMoveable); }
     bool blockProjectile() { return m_attribs.has(ThingAttrBlockProjectile); }
     bool isNotPathable() { return m_attribs.has(ThingAttrNotPathable); }
+	void setPathable(bool var);
```
edubart: Wrong indentation, also not best place to add the function (in the middle of isXXX() functions)

EgzoT: Thanks for advice.
I moved function to the additional section.
Fixed indentation.

PR #905, opened 2017-11-16
Attempt to fix #852

### LINKED ISSUE #852: [Missing Feature / Bug] Dat Outfit Group Flag
### Before creating an issue, please ensure:
- [X] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/otclient.494/ for support)
- [X] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Select the outfit 724/725 (Evoker) and use the bat addon.
image

### Expected behaviour
The outfit should be reproducing the idle/stand animation.

### Actual behaviour
The outfit is not reproducing any animation when is idle/stand.
Also, maybe for this reason, the walking animation is not using the correct sprites.

### Environment
Protocol 10.99 with latest OTClient.

### More informations
I don't know exactly when the Tibia added the group animation flag on the dat, but it works like this:
- You can have 2 groups of animations: 1. Idle/Stand and the 2. Walking.
- You can set different values of width, height, crop size, layers, X, Y, Z and animations count for each one animation group. It means I could do like the Evoker outfit: make the bat animated while stand, for example, with maybe like 10 animations and size like 3x3 and, on the walking animation, with just 4 animations and size like 2x2.

> Also, I tried to use the outfits 24 (Minotaur Archer), 81 (Cobra), 274 (Squirrel) outfits. Which each one have the idle animation with size 1x1 of sqm and animation 1, but on the group 2 (animated) they all have sizes 2x2 of sqms and more than 1 animation. On the game it crashes when these outfits moves.
I reported this once here ( https://github.com/edubart/otclient/issues/813 ), but I found a "fix" just setting the idle/stand animation size to 1x1 and filling the sprites correctly with this new size. It didn't crashed anymore until today. But... this is not a fix, right? The group flag is not working properly as it should, I think.

ninjalulz:
Quite possible that it breaks backwards compatibility though.

### DIFF
--- a/src/client/creature.cpp
+++ b/src/client/creature.cpp
@@ -107,12 +107,12 @@ void Creature::internalDrawOutfit(Point dest, float scaleFactor, bool animateWal
 
     // outfit is a real creature
     if(m_outfit.getCategory() == ThingCategoryCreature) {
-        int animationPhase = animateWalk ? m_walkAnimationPhase : 0;
 
-        if(isAnimateAlways() && animateIdle) {
-            int ticksPerFrame = 1000 / getAnimationPhases();
-            animationPhase = (g_clock.millis() % (ticksPerFrame * getAnimationPhases())) / ticksPerFrame;
-        }
+        int animationPhase = animateWalk ? m_walkAnimationPhase : 0;
+        if(getIdleAnimator() != nullptr && !isWalking() && animateIdle)
+            animationPhase = getIdleAnimator()->getPhase();
+        else if(getAnimator() != nullptr && animateWalk && isWalking())
+            animationPhase = m_walkAnimationPhase;
 
         // xPattern => creature direction
         int xPattern;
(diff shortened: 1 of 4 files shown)


edubart:
I personally can't test this at the moment, I can merge if more people test and approve

edubart:
I'm gona also note that this need careful test across different client versions, because its changing a critical part of the animation code

src/client/creature.cpp
```diff
@@ -107,12 +107,12 @@ void Creature::internalDrawOutfit(Point dest, float scaleFactor, bool animateWal
...
     // outfit is a real creature
     if(m_outfit.getCategory() == ThingCategoryCreature) {
-        int animationPhase = animateWalk ? m_walkAnimationPhase : 0;
 
-        if(isAnimateAlways() && animateIdle) {
-            int ticksPerFrame = 1000 / getAnimationPhases();
-            animationPhase = (g_clock.millis() % (ticksPerFrame * getAnimationPhases())) / ticksPerFrame;
-        }
+        int animationPhase = animateWalk ? m_walkAnimationPhase : 0;
+        if(getIdleAnimator() != nullptr && !isWalking() && animateIdle)
```
edubart: You completely ignored the `isAnimateAlways` so I guess this will cause regression in something version

PR #906, opened 2017-11-16
Remove the dumper class

diath:
This PR removes a class that wasn't really used anywhere and doesn't seem to serve much purpose when one can just use the templated print function instead to dump variables.


edubart:
I personally use that class to quickly debug variables in C++, like this
```C++
dump << something << otherThing;
```
and it's not harming anything

PR #907, opened 2017-11-16
Add a keyboard shortcut to toggle topmenu visibility

diath:
This PR adds a shortcut (Ctrl-Shift-T) that lets you toggle the top menu visibility to save some vertical space.


edubart:
I believe this would also need a small button on the top when the top menu is hidden to show the top menu again, in case user "miss clicked" the hotkey and he doesn't know how to recover

EgzoT:
Maybe something like that (this icons so that big, because I maked these icons for small touch screens).

You must add to:
```
mapPanel = modules.game_interface.getMapPanel()
backgroundPanel = modules.client_background.getBackground()
```

show_hide_button_2_17 11 2017
show_hide_button_17 11 2017_333

Issue #75, opened 2012-08-20
Create a mana bar beneath players health bar


BenDol:
Should provide the option to display a mana bar beneath the players health bar.

Felipe1993:
any tutorial regarding how to use this feature?

Felipe1993:
i had edited the source for a mana baar now i need the script to display it
any idea?

TheSumm:
Implemented in #909.

Showing the current mana of all players is not possible at this point due to game server limitations. As I have not seen the feature on any server, I would also argue that it is not common enough to warrant adding it to the client at this point.

PR #909, opened 2017-11-16
Add player mana bar

diath:
This partially addresses #75. It merely adds a mana bar for just the local player though and not all players but it can be easily extended with custom op codes if needed.


andersonfaaria:
+1 for making a option to turn it on/off 👍

Issue #903, opened 2017-11-15
Missing delay for some actions


andersonfaaria:
There's some cases where we definitely should have a delay to control the actions, principally if the player isn't a game master. One of the things that always piss me off is the CTRL + ARROW key to change direction. 
![tibia](https://user-images.githubusercontent.com/8560144/32859691-5dfbb2a0-ca36-11e7-8180-3e5e8dbd370c.gif)
This happens because the game is too sensible with the arrow keys and when you press CTRL + -> it interprets hundreds of commands instead of pressing one time and adding a delay to the next action;

### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/otclient.494/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Hold Ctrl + /\ (up arrow key)
2. While still holding Ctrl Switch to -> (left arrow key)
3. If you do this rapidly or hold for enough time, the movement won't be fluid and a glitch will occur.

### Expected behaviour

The game should have a way to filtering when you are holding a key. Perhaps ignoring holding for most of the actions except hotkey and for the hotkeys we should have at least a safe delay like 100ms or something like that.

### Actual behaviour

No delay at all, if you hold a button you are catched on the maxPacketsPerSecond configuration of tfs and the server disconnects you.

soul4soul:
This seems related to issue #564

andersonfaaria:
this one was specific about hotkeys, but definitely related. 
I think we should have a class that filters every action of the game and decide if the action should pass or be filtered. This would be a huge improvement.
Something like 'processAction'

Issue #564, opened 2014-09-06
Limiting packets sent by OTClient


Sceptero:
Many new servers use "maxPacketsPerSecond = 25". Holding down a hotkey button pressed results in a kick from those servers. Cipsoft's tibia client must be limiting number of packets which are sent while holding down a hotkey because you can do it and not get kicked out of the server. Maybe adding some configurable delay between each packet of the same type would solve the problem.

gesior:
Code to limit hotkey (not packets per second of whole client).
In source **src\framework\luafunctions.cpp** add new function (we need high precision time in LUA):
    
    g_lua.bindGlobalFunction("timeMillis", []() { return (int) stdext::millis(); });
    
In modules **game_hotkeys\hotkeys_manager.lua** before:

    function doKeyCombo(keyCombo)

Add:

    local lastHotkeyUse = 0

After:

    if hotKey.autoSend then

Add:

    if timeMillis() - lastHotkeyUse < 100 then return end
    lastHotkeyUse = timeMillis()

Simple limit until we get source fix for it. 100 = 100 miliseconds (0.1 sec) between hotkeys actions.
Tested on newest sources.

PR #910, opened 2017-11-16
Add turn and hotkey delay

diath:
This partially addresses #903 and #564.

I don't think limiting the packet rate is a good solution since the behavior could be desired in many actions and would otherwise make the client feel clunky but since hotkey actions are sent at the rate of foreground FPS this should address the issue where too many actions are being sent by turn and hotkey keys.

The delay can be configured through Options -> Game settings panel.

### DIFF
--- a/modules/client_options/game.otui
+++ b/modules/client_options/game.otui
@@ -38,11 +38,51 @@ Panel
     id: displayText
     !text: tr('Display text messages')
 
+  Label
+    id: turnDelayLabel
+    !text: tr('Turn delay: %sms', 30)
+    anchors.left: parent.left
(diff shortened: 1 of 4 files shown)


andersonfaaria:
Perhaps you can answer this better than me but by holding the arrow keys to walk is it spamming as well? should we made this solution to walking as well? Could it be the cause of the wrong animation while walking?

diath:
I believe it's already being limited based on the player's walking speed:
https://github.com/edubart/otclient/blob/master/src/client/game.cpp#L599

edubart:
Nice work @diath

PR #917, opened 2017-11-21
Wrong call of child instead of parent

andersonfaaria:
showTopMenuButton is calling a child, returning syntax error and making the client unable to open. This change will fix this issue.


diath:
The button is a child of mapPanel, not rootPanel: https://github.com/edubart/otclient/blob/47766685d9f28fdbaeefbf89095dc6060a351c1b/modules/game_interface/gameinterface.otui#L35
Perhaps you were trying to apply the changes by hand and forgot an indentation level?

andersonfaaria:
makes sense, I'll check it when I get home and answer with the result.

PR #918, opened 2017-11-28
Use GMP or OpenSSL for RSA

JSkalskiSBG:
Not tested on machine with OpenSSL. I don't have any PC with working OpenSSL library.
Tested with GMP:

    cmake -DUSE_GMP=1 ..

USE_GMP=0 or no parameter will compile with OpenSSL:

    cmake -DUSE_GMP=0 ..
    
    cmake ..

I tried to detect what version of OpenSSL is loaded by cmake, but could not find that information.

Refs #837


JSkalskiSBG:
Added fix for RSA formula in GMP implementation.

diath:
I think a cleaner approach for dual library support would be two implementation files and choosing the correct one via CMake based on the USE_GMP flag rather than littering the code with ifdefs but that's just my opinion.

JSkalskiSBG:
@diath 
And what about not implemented functions in crypt.cpp without RSA (sha, md5, generate key)? Leave them and show some error in console ('Compile with OpenSSL to use') or add in other files '#ifdef USE_GMP' ?

diath:
I would say yes, just leave empty implementation and issue a warning. You should probably wait until someone else weighs in though because I don't know if that's how @edubart would like it to be implemented.

JSkalskiSBG:
@diath 
Others may like it or not, but with my last commit it's proper OOP code.
Crypt class methods (hash and RSA) call methods in new class CryptGMP or CryptOpenSSL. UUID related methods (same for GMP and OpenSSL) are still in Crypt class.
Calling not implemented method from LUA shows error in console.

**Can anyone try to compile it with OpenSSL (no param to cmake)?** I don't have machine to test it.
**Can anyone update/prepare project for Visual Studio?** I don't have working version and I got no idea about VS.

conde2:
While you're at this maybe you could update the code to use the newer OpenSSL version ?

JSkalskiSBG:
@conde2 
Ok. I wrote version for OpenSSL 1.1
Now I need to make it compatible with 1.0 and 1.1, but there is a tutorial ( https://wiki.openssl.org/index.php/OpenSSL_1.1.0_Changes ) , so it should not be hard.

// gesior at work

JSkalskiSBG:
Tested **rsaEncrypt** and **rsaDecrypt** with **OpenSSL 1.1.0g** and **GMP 6**.
If you compile with OpenSSL, you can use hash functions, **g_crypt.rsaGenerateKey(bitsLength, e)** and **g_crypt.rsaCheckKey()**

To compile with OpenSSL:

    cmake ..

or:

    cmake -DUSE_GMP=0 ..

To compile with GMP:

    cmake -DUSE_GMP=1 ..

OpenSSL documentation says, that 'compability code' I've added, should make it work with OpenSSL 1.0.2+ and 1.1.0.

# **Can anyone test it with OpenSSL on 1.0.x?**

iryont:
Since we have compatibility for OpenSSL < 1.1.0 and >= 1.1.0 now I believe this PR is no longer needed. If anyone disagrees let me know, so we can discuss it.

PR #925, opened 2017-12-05
Added new health bar styles

Ceetros:


diath:
This is inefficient, the texture will be loaded with every frame drawn, you should store it as a class member.

hrsantiago:
And also should follow the example of skull/emblems/shields to set the filename

Ceetros:
Yeah, im updating this, a momment :)

EgzoT:
I think better way is modify this to choose type of healthbar style in lua:

- Without image (old style)

- With image (new style)

Riverlance:
I tried it here, and it lags too much.
The drawInformation function executes too much times and quickly, so **your FPS will decrease significantly**.

My solution here was to create the texture as a creature variable, not as you done inside the drawInformation.

At `Creature::Creature()`, you add like this:
``` C++
m_creatureBarTexture = g_textures.getTexture("/images/game/lifebar/background");
```

Now you are free to use this texture on drawInformation without lag.

PR #929, opened 2017-12-28
Update translations

vamo89:
Update pt and pl translations


conde2:
It would be better if you remove the ":" from the tr() functions, because ":" are not supposed to be translated.

vamo89:
Thanks for the info. I've updated some, but there's some more to go. I'll add more until it's done for pt translations.

PR #932, opened 2017-12-31
Add setBuffer() method to OutputMessage class

EgzoT:
This method similar to method setBuffer of InputMessage class will able to easy data copy from one OutputMessage/InputMessage to other.

May be helpful for make cast system using TCP/IP protocol.

### DIFF
--- a/src/framework/net/outputmessage.cpp
+++ b/src/framework/net/outputmessage.cpp
@@ -37,12 +37,12 @@ void OutputMessage::reset()
 
 void OutputMessage::setBuffer(const std::string& buffer)
 {
-	int len = buffer.size();
-	checkWrite(MAX_HEADER_SIZE + len);
-	memcpy(m_buffer + MAX_HEADER_SIZE, buffer.c_str(), len);
-	m_writePos = MAX_HEADER_SIZE;
-	m_headerPos = MAX_HEADER_SIZE;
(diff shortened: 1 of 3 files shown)


src/framework/net/outputmessage.h
```diff
@@ -40,6 +40,7 @@ class OutputMessage : public LuaObject
 
     void reset();
 
+	void setBuffer(const std::string& buffer);
```
ubap: incorrect identation.
tab -> spaces

src/framework/net/outputmessage.cpp
```diff
@@ -35,6 +35,16 @@ void OutputMessage::reset()
     m_messageSize = 0;
 }
 
+void OutputMessage::setBuffer(const std::string& buffer)
+{
+	int len = buffer.size();
```
ubap: incorrect identation.
tabs -> spaces

EgzoT:
@ubap Thx, for catch this.
In fact I always forget about indentation, I must remember about indentations in future.

hrsantiago:
@EgzoT however you PR now is a mess :disappointed: would you mind sending a new one with only one commit?

EgzoT:
@hrsantiago I added New push request, with one commit.
#967

PR #939, opened 2018-01-28
Load appropiate fonts charset to language (support polish signs)

EgzoT:
It's provide to support polish signs [ąĄęĘśŚćĆżŻźŹóÓ] in some fonts (cp1250) (maybe new languages signs in future) when user choose polish language.
Now only polish language have other charset (cp1250), other languages use cp1252, when language was change, charset will be reload to right charset (cp1252/cp1250). 

Read charset from locales files (locales.charset), and load fonts files from folder (charset name), other/universal fonts will be load like in previous time.


hrsantiago:
A true type font support would fit better here

EgzoT:
Probably yes, but now we don't have true font support. Some people search for this, so I have make this. Maybe this is not best solution, but work good.

slavidodo:
I think a proper approach used by many games - and loyalty game engines - is to render the textures in a big atlas containing a predefined charset, say ASCII.

We could use TTF and cache the texture somewhere in the LOCAL files, this will cause some overhead probably when opening the client for the first time but it will have the same performance as Bitmap anyways.

Issue #468, opened 2014-02-05
Settings aren't saved(windows 7)


Zbizu:
as in title - administrator account name on my pc contains "Ł" letter and there is no way to load/save settings and minimap (some files are created, but they are never loaded again)

can these be saved in client directory as default path?

software: windows 7 home 64 bit

kungen32:
> Otclient made empty file called otclient instead of creating a new folder.
> Replacing empty file to new folder solves this problem.
> I'll leave this issue open till someone fix it.

Hello

Can you be more simplistic in your answer ?
I just found c:users:otclient, contains 2 files, not emptys
Also %appdata% otclient/otclient/otclient ( 3 times ) then a config.

Where i remplace the file for the new one, and which file ?

Zbizu:
The issue was with my user name containing an unhandled character probably (Ł).

PR #947, opened 2018-02-23
Fix write dir missing a trailing slash

diath:
Newer PhysFS version requires a trailing slash for the write directory, otherwise an empty file is created instead.
Older versions should remain unaffected.

This fixes https://github.com/edubart/otclient/issues/468#issuecomment-36684359.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -22,7 +22,7 @@ end
 g_resources.addSearchPath(g_resources.getWorkDir() .. "mods", true)
 
 -- setup directory for saving configurations
-g_resources.setupUserWriteDir(g_app.getCompactName())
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -22,7 +22,7 @@ end
 g_resources.addSearchPath(g_resources.getWorkDir() .. "mods", true)
 
 -- setup directory for saving configurations
-g_resources.setupUserWriteDir(g_app.getCompactName())
+g_resources.setupUserWriteDir(('%s/'):format(g_app.getCompactName()))
```
marmichalski: IMHO it would be better to handle it in sources.

diath: That would be just working around a misuse of the library.

Issue #624, opened 2015-01-14
Warnings compiling on Mac OS X and fatal error trying to run client


richigarza:
after problems with lua library, i get this warnings:

In file included from /Users/richigarza/otclient/src/framework/net/outputmessage.cpp:24:
In file included from /Users/richigarza/otclient/src/framework/../framework/util/crypt.h:29:
In file included from /usr/local/include/boost/uuid/uuid.hpp:203:
/usr/local/include/boost/uuid/detail/uuid_x86.hpp:42:5: warning: 'register'
      storage class specifier is deprecated [-Wdeprecated-register]
    register __m128i mm = uuids::detail::load_unaligned_si128(data);
    ^~~~~~~~~
/usr/local/include/boost/uuid/detail/uuid_x86.hpp:53:5: warning: 'register'
      storage class specifier is deprecated [-Wdeprecated-register]
    register __m128i mm_this = uuids::detail::load_unaligned_si128(data);
    ^~~~~~~~~
/usr/local/include/boost/uuid/detail/uuid_x86.hpp:54:5: warning: 'register'
      storage class specifier is deprecated [-Wdeprecated-register]
    register __m128i mm_rhs = uuids::detail::load_unaligned_si128(rhs.data);
    ^~~~~~~~~
/usr/local/include/boost/uuid/detail/uuid_x86.hpp:61:5: warning: 'register'
      storage class specifier is deprecated [-Wdeprecated-register]
    register __m128i mm_left = uuids::detail::load_unaligned_si128(lhs.data);
    ^~~~~~~~~
/usr/local/include/boost/uuid/detail/uuid_x86.hpp:62:5: warning: 'register'
      storage class specifier is deprecated [-Wdeprecated-register]
    register __m128i mm_right = uuids::detail::load_unaligned_si128(rhs.data);
    ^~~~~~~~~
/usr/local/include/boost/uuid/detail/uuid_x86.hpp:64:5: warning: 'register'
      storage class specifier is deprecated [-Wdeprecated-register]
    register __m128i mm_cmp = _mm_cmpeq_epi32(mm_left, mm_right);
    ^~~~~~~~~
/usr/local/include/boost/uuid/detail/uuid_x86.hpp:74:5: warning: 'register'
      storage class specifier is deprecated [-Wdeprecated-register]
    register __m128i mm_left = uuids::detail::load_unaligned_si128(lhs.data);
    ^~~~~~~~~
[...]

Issue #617, opened 2015-01-01
SoundBuffer Error OSX


gpedro:
I'm using OSX and get this error after login

	ERROR: Unable to send extended opcode 1, extended opcodes are not enabled
	Assertion failed: (alGetError() == AL_NO_ERROR), function ~SoundBuffer, file /Users/gpedro/otclient/src/framework/sound/soundbuffer.cpp, line 36.

gpedro:
Duplicate: #555

gpedro:
Workaround:

```diff
diff --git a/modules/client/client.lua b/modules/client/client.lua
index eb7a81a..a3c463f 100644
--- a/modules/client/client.lua
+++ b/modules/client/client.lua
@@ -27,12 +27,12 @@ end

 function startup()
   -- Play startup music (The Silver Tree, by Mattias Westlund)
-  musicChannel:enqueue(musicFilename, 3)
-  connect(g_game, { onGameStart = function() musicChannel:stop(3) end })
-  connect(g_game, { onGameEnd = function()
-      g_sounds.stopAll()
-      musicChannel:enqueue(musicFilename, 3)
-  end })
+  -- musicChannel:enqueue(musicFilename, 3)
+  -- connect(g_game, { onGameStart = function() musicChannel:stop(3) end })
+  -- connect(g_game, { onGameEnd = function()
+  --     g_sounds.stopAll()
+  --     musicChannel:enqueue(musicFilename, 3)
+  -- end })

   -- Check for startup errors
   local errtitle = nil
```

Issue #555, opened 2014-08-14
Assertion failed: (alGetError() == AL_NO_ERROR), function ~SoundBuffer


jollejkpg:
Assertion failed: (alGetError() == AL_NO_ERROR), function ~SoundBuffer, file /otclient-master/src/framework/sound/soundbuffer.cpp, line 39.

(Client) crashreport.txt
== application crashed
app name: OTClient
app version: 0.0.2
build compiler: gcc 4.8.1
build date: Aug 11 2014
build type: RelWithDebInfo
build revision: 0 (Red Eyes)
crash date: Aug 15 2014 00:06:24
exception: Access violation (0xc0000005)
exception address: 0x0068bea9
  backtrace:
    0: C:\Client\OTClient.exe(wglewIsSupported+0xc2b9) [0x000000000068BEA9]
    1: C:\Client\OTClient.exe [0x00000000004EE71F]
    2: C:\Client\OTClient.exe [0x00000000004EF636]
    3: C:\Client\OTClient.exe [0x00000000004C7A63]
    4: C:\Client\OTClient.exe [0x00000000004C8068]
    5: C:\Client\OTClient.exe [0x00000000004ED6DC]
    6: C:\Client\OTClient.exe(wglewIsSupported+0xd8f3e) [0x0000000000758B2E]
    7: C:\Client\OTClient.exe(wglewIsSupported+0x12dbbc) [0x00000000007AD7AC]
    8: C:\Client\OTClient.exe(wglewIsSupported+0xde72a) [0x000000000075E31A]
    9: C:\Client\OTClient.exe [0x0000000000474595]
    10: C:\Client\OTClient.exe(alGetListeneriv+0x4007b) [0x0000000000564C1B]
    11: C:\Client\OTClient.exe(alGetListeneriv+0x41434) [0x0000000000565FD4]
    12: C:\Windows\syswow64\msvcrt.dll(malloc+0x57) [0x0000000075A69D45]
    13: C:\Client\OTClient.exe(alGetListeneriv+0x40500) [0x00000000005650A0]

After being online for a while and walking around the client suddenly crashes, giving this error.

Issue #340, opened 2013-07-28
Assertion failed: (alGetError() == AL_NO_ERROR)


marksamman:
Assertion failed: (alGetError() == AL_NO_ERROR), function ~SoundBuffer, file /Users/mark/Projects/otclient/src/framework/sound/soundbuffer.cpp, line 36.

It happens a few seconds after logging in to gameworld.

Operating system: Mac OS X 10.8.4

PR #953, opened 2018-03-11
Dodge MacOS crash, fix battle, remove flagwindow.otui (not used)

4drik:
Temp solution to issue #340, #555, #617, #624 (we should find better).
Has someone plan to use flagwindow.otui? (removed for now)


EgzoT:
1) But why you delete flagwindow? How we will may now add flag to map?
2) For check OS name OTClient have inbuilt method: g_platform.getOSName().
3) Why you block music for others OS than Windows? I tested it on Android and it's works. Probably in desktop linux too (but I don't tested that).

4drik:
1. But why you delete flagwindow? How we will may now add flag to map?
We are currently using minimap flag from the 30-minimap.otui file.
2. Thank you. I will use it.
3. I tested it on MacOS and Windows, so I was not sure how about other systems.
Gonna fix it too, thank you.

diath:
I think that instead of checking for the OS to disable music you should add a global Lua flag that has a different value depending whether the sound framework is present or not and don't play audio based on that so then you can just disable it with -DFRAMEWORK_SOUND=0 since currently compiling without sound will make modules not load so that would fix this issue and also let people on macOS disable it to prevent the bug from happening if needed.

PR #965, opened 2018-04-23
Fix tab blink

diath:
The event should be removed when selecting a tab, otherwise when you select a blinking tab and then deselect it, it will remain in the urgent state.


iryont:
Good job.

PR #967, opened 2018-04-24
Add setBuffer() method to OutputMessage class

EgzoT:
This method similar to method setBuffer of InputMessage class will able to easy data copy from one OutputMessage/InputMessage to other.

May be helpful for make cast system using TCP/IP protocol.


EgzoT:
I added new pull request to avoid mess.

iryont:
I will merge this PR, but it requires a few fixes which I will implement right after the merge since I cannot do it within PR due to removed source of the PR (unknown repository).

PR #974, opened 2018-06-16
Use gameLeftPanel as gameRightPanel

EgzoT:
It's provide to use gameLeftPanel (if it's ON) like gameRightPanel, by using function addToPanels().

When gameRightPanel is full:
leftpanel_as_rightpanel_16-06-2018

When in gameRightPanel one window is resized (now windows it's moved to gameLeftPanel, normal it's would be removed):
leftpanel_as_rightpanel_16-06-2018_2

**New functions:**
[modules.game_interface] **addToPanels(`<UIWidget>`)** - they set uiminiwindow in gameRightPanel or in gameLeftPanel (according on free place) priority gameRightPanel [function may set uiminiwindow to gameLeftPanel only when it's ON], additional they moved (if find free place in gameRightPanel or gameLeftPanel) window (instead remove) to other Panel.

UIMiniWindowContainer:**getEmptySpaceHeight()** - return height of UIMiniWindowContainer without windows (uiminiwindow), who we may set to other windows.

**New signals (signalcall):**
UIMiniWindowContainer.**onRemoveFromContainer [`<UIWidget>`]** - return UIWidget who want to be removed (before deleted who provide to set operation to stopped deleted this UIWidget [change parent]) from UIMiniWindowContainer (by UIMiniWindowContainer:fitAll()).

### DIFF
--- a/modules/corelib/ui/uiminiwindowcontainer.lua
+++ b/modules/corelib/ui/uiminiwindowcontainer.lua
@@ -9,6 +9,20 @@ function UIMiniWindowContainer.create()
   return container
 end
 
+function UIMiniWindowContainer:getEmptySpaceHeight()
+  local sumHeight = 0
+  local children = self:getChildren()
+
+  for i=1,#children do
+    if children[i]:isVisible() then
+      sumHeight = sumHeight + children[i]:getHeight()
+    end
+  end
+  local selfHeight = self:getHeight() - (self:getPaddingTop() + self:getPaddingBottom())
+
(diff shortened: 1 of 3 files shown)


Felipe21:
I have extra right panel on my otclient how can i add that panel into this code?

slavidodo:
This is a good start at which we can actually support multiple containers.
Any chance we could see this dynamically handled by the user using an "arrow" on the top corner of the last panel?

I'm looking to merging this!

PR #981, opened 2018-07-15
Compilation & linking fixes

DSpeichert:
* Compilation fixes (see https://github.com/otland/forgottenserver/commit/7193a670c6b5b7cff472e1946e5b1e7ca7e1051b#diff-70002a4cfb8e3fd5d00493262492a5daR27)
* OpenAL path compatible with vcpkg
* Optimizations in MSVS

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,8 @@ Thumbs.db
 .directory
 src/framework/graphics/dx/
 modules/.project/modules.sublime-workspace
+vc14/.vs/
+*.dll
 
 #################
 ## Visual Studio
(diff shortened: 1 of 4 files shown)


diath:
* The changes in connection.h should be indented with spaces, not tabs.
* Does the EnableFiberSafeOptimizations even have any effect on OTC? Based on what I could gather from MSDN it's not really going to affect anything.
* I don't think AVX should be enabled by default since unlike server software the client is meant to be redistributed in which case we should be aiming for the most common denominator plus any potential performance benefits, while losing platform support, will be dwarfed by bottlenecks in the graphics engine.

DSpeichert:
@diath 
* Fixed formatting
* I am not completely sure if EnableFiberSafeOptimizations will change anything but it won't hurt when it doesn't apply. Similar set of optimizations are now used in TFS.
* AVX is available in all processors released since 2011, which is quite common by now. I didn't choose the AVX2 which would require last year's models.

PR #986, opened 2018-08-05
Function updateJump() fix

okk3:
nextT should cast double as it's a second degree equation and results in a double type. -> Bhaskara.

### DIFF
--- a/src/client/creature.cpp
+++ b/src/client/creature.cpp
@@ -405,7 +405,7 @@ void Creature::updateJump()
 
         int nextT, i = 1;
         do {
-            nextT = stdext::round((-b + std::sqrt(std::max<int>(b*b + 4*a*(roundHeight+diff*i), 0.0)) * diff) / (2*a));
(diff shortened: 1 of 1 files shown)


EgzoT:
Maybe someone want to describe, how to reproduce this issue, because I don't have problem/don't use jump function.
That information may be helpful for contributors, who want to test this fix and merge.

Riverlance:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/otclient.494/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Login into any character on your server (doesn't needs to be a Game Master)
2. Open your client's terminal window
3. Execute the following commands:
``` LUA
local localPlayer = g_game.getLocalPlayer()
localPlayer:jump(32, 1000) -- Jumping 32 pixels in 1000 ms
```

### Expected behaviour
The player should go up and then go down, in 1000ms total.

### Actual behaviour
The player goes up, but, at the moment that the player should go down, the client crashes.

### Environment
OTClient Latest

### Fix by @okk3 
The cast should return a double value, instead int.

PR #987, opened 2018-08-21
Add -no-pie option to CMAKE_CXX_LINK_FLAGS

ltgouvea:
Fixes compiling issues with libogg in Mint Linux 19
----
This is the error I've encountered compiling otclient in Mint Linux 19 without this option
```
/usr/bin/x86_64-linux-gnu-ld: /usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/libogg.a(framing.o): relocation R_X86_64_32S against `.rodata' can not be used when making a PIE object; recompile with -fPIC
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ include(src/client/CMakeLists.txt)
 
 # functions map for reading backtraces
 if(NOT APPLE)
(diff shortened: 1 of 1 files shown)


JSkalskiSBG:
I had to edit this line to:
```
set(CMAKE_CXX_LINK_FLAGS "${CMAKE_CXX_LINK_FLAGS} -fPIC -no-pie -Wl,-map=${PROJECT_NAME}.map")
```

to make it work.

PR #928, opened 2017-12-27
Fix compiling error on ubuntu 17.10

vamo89:
Using shared OGG library instead of static OGG library to fix this problem:
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/libogg.a(framing.o): relocation R_X86_64_32S against .rodata' can not be used when making a shared object; recompile with -fPIC /usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/libogg.a(bitwise.o): relocation R_X86_64_32S against.rodata' can not be used when making a shared object; recompile with -fPIC

I'm not sure if this is the right way to fix, but worked for me.
This are the libraries I'm using:
```
||/ Name                      Version                     Architecture
+++-=========================-===========================-============
ii  build-essential           12.4ubuntu1                 amd64       
ii  cmake                     3.9.1-1                     amd64       
ii  git-core                  1:2.14.1-1ubuntu4           all         
ii  libboost-all-dev          1.62.0.1                    amd64       
ii  libglew-dev:amd64         2.0.0-5                     amd64       
ii  liblua5.1-0-dev:amd64     5.1.5-8.1build1             amd64       
ii  libopenal-dev:amd64       1:1.17.2-4                  amd64       
ii  libphysfs-dev             2.0.3-6                     amd64       
ii  libssl-dev:amd64          1.0.2g-1ubuntu13.3          amd64       
ii  libvorbis-dev:amd64       1.3.5-4                     amd64       
ii  zlib1g-dev:amd64          1:1.2.11.dfsg-0ubuntu2      amd64       
```

and besides that, I've biuld physfs from stable 3.0 sources

PR #1001, opened 2018-12-05
OpenSSL >= 1.1.0 compatibility

divinity76:
OpenSSL did some otclient-breaking changes around 1.1.0, 
this patch should work on both < and >= 1.1.0

- **warning**: UNTESTED as of writing. this fixes the compilation errors, but since i only work on tibia 7.6 (which doesn't use encryption, encryption was implemented in tibia 7.7), i haven't actually tested this patch, should probably get someone to test it before pulling it upstream.

### DIFF
--- a/src/framework/util/crypt.cpp
+++ b/src/framework/util/crypt.cpp
@@ -335,7 +335,7 @@ void Crypt::rsaSetPublicKey(const std::string& n, const std::string& e)
         m_rsa->_method_mod_n = nullptr;
     }
 #else
-    BIGNUM *bn, *be;
+    BIGNUM *bn = nullptr, *be = nullptr;
     BN_dec2bn(&bn, n.c_str());
     BN_dec2bn(&be, e.c_str());
     RSA_set0_key(m_rsa, bn, be, nullptr);
@@ -360,7 +360,7 @@ void Crypt::rsaSetPrivateKey(const std::string& p, const std::string& q, const s
         m_rsa->_method_mod_q = nullptr;
     }
 #else
-    BIGNUM *bp, *bq, *bd;
+    BIGNUM *bp = nullptr, *bq = nullptr, *bd = nullptr;
     BN_dec2bn(&bp, p.c_str());
     BN_dec2bn(&bq, q.c_str());
     BN_dec2bn(&bd, d.c_str());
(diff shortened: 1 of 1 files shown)


iryont:
I haven't checked whether the code works either, but I did fix the code style to fit OTClient one.

I do not have OpenSSL >= 1.1.0 myself, so I cannot really merge this branch before someone tests it and confirms it's good to go.

src/framework/util/crypt.cpp
```diff
@@ -327,60 +327,45 @@ void Crypt::rsaGenerateKey(int bits, int e)
...
 #else
-	{
-		BIGNUM *bn=NULL;
-		BIGNUM *be=NULL;
-		BN_dec2bn(&bn, n.c_str());
-		BN_dec2bn(&be, e.c_str());
-		RSA_set0_key(m_rsa,bn,be,NULL);
-		// note, not supposed to free bn/be here, that's m_rsa's destructor's job
-	}
+    BIGNUM *bn, *be;
```
divinity76: error: bn and be is not guaranteed by c++ specs to be null-initialized here, and if they're not, the BN_dec2bn() call below will try to read arbitrary memory and UB ensues (most likely it would crash for trying to read uninitialized memory, but it could do other stuff too, if they accidentally pointed to a valid allocated memory location)

- compilers *MAY* null-initialize these variables just for the heck of it, but the c++ specs does not guarantee it, AND null-initializing memory *DOES* incur a runtime performance hit, so a compiler that is instructed to optimize the run-time code for speed should not do it...

the code should be ```BIGNUM *bn=nullptr, *be=nullptr;```  or something like that.

iryont: It appears that BN_dec2bn does in fact re-use memory if it's already initialized. I thought it doesn't and it just creates a new BIGNUM instead, so yea, default value of nullptr is required.

divinity76: that looks safer, thanks :)

alvelvis:
HI!
I had been struggling to compile because of the OpenSSL 1.1.x problem, so I downloaded divinity76:patch-2 branch and it compiled after a small problem: in the end it printed an error and asked me to recompile with -fPIC parameter. After following instruction in #928. it compiled just fine. Thank you! ;) Any further information you might want to know about the process please just let me know (and how to retrieve it). I've been using OTC in a 8.60 OT serv, Ubuntu 18.04.

iryont:
Lovely :) In such case I will assume it is working properly and I will merge it. Thank you.

PR #1002, opened 2019-01-05
Fix magic effect animations in newer client versions

diath:
This makes the magic effects use their animation frame duration based on the animation data from the .dat file instead of a static duration like older client versions.


iryont:
In general this whole Animator class should be optimized a bit more. Too much use of loops inside functions which are called each time something is drawn. I mean even getTotalDuration uses "for" which isn't good. Such things can be easily optimized.

I will merge your code though. Good job.

PR #1005, opened 2019-01-05
Don't apply message prefixes to messages with empty name

diath:
This suppresses message prefixes on messages without a player name as those are often used for event announcements and result in a weird message format.


TheSumm:
Thanks

PR #1006, opened 2019-01-05
Improve scrollbar navigation

diath:
This adds extra convenience key modifiers to the scrollbar controls for moving to the beginning/end and by -10/+10.


iryont:
I guess it might be useful in general and it doesn't hurt, so merged and good job.

PR #1008, opened 2019-01-30
Fix glew32/zlib debug build.

Arkamek:


iryont:
I doubt this will be useful in general. The naming purely depends on the library itself and you didn't change other libraries with a suffix of d. Also, debug mode in OTClient doesn't automatically mean you want to link against debug libraries as well. You are usually looking for bugs within OTClient code rather than third party libraries, thus linking against debug libraries is not mandatory in debug mode.

Unless I'm missing something I have always linked against normal libraries even with a debug mode.

diath:
@iryont 
>I doubt this will be useful in general. The naming purely depends on the library itself and you didn't change other libraries with a suffix of d. 

This is true but the default output file for those two libraries in debug mode contains the d suffix whether you build from the source or using vcpkg when using MSVC as the compiler so while you can have arbitrary names it makes sense to use the default ones.


>Also, debug mode in OTClient doesn't automatically mean you want to link against debug libraries as well. You are usually looking for bugs within OTClient code rather than third party libraries, thus linking against debug libraries is not mandatory in debug mode.

This can however lead to runtime library incompatibility.

iryont:
I have mixed feelings, but I will merge it nonetheless. It's hardly a problem for any programmer to change it when he wants, but if it will help newcomers I'm fine with it.

PR #1021, opened 2019-03-09
Support for OpenSSL with a fallback to GMP

iryont:
This is a new code based on posted PRs. However, it uses automatic fallback to GMP if OpenSSL is not found.

I did not test it myself. Also, the implemention is open for discussion.

### DIFF
--- a/src/framework/CMakeLists.txt
+++ b/src/framework/CMakeLists.txt
@@ -219,14 +219,12 @@ endif()
 message(STATUS "LuaJIT: " ${LUAJIT})
 
 find_package(PhysFS REQUIRED)
-find_package(OpenSSL REQUIRED)
 find_package(ZLIB REQUIRED)
 
 set(framework_LIBRARIES ${framework_LIBRARIES}
     ${Boost_LIBRARIES}
     ${LUA_LIBRARY}
     ${PHYSFS_LIBRARY}
-    ${OPENSSL_LIBRARIES}
     ${ZLIB_LIBRARY}
 )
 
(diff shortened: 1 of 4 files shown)


iryont:
Another idea worth implementing - we never need to set d. It can be calculated using p, q and e. The same goes to rsaSetPrivateKey which should set both public and private keys (p, q, n, d and e).

That being the case having p, q and e is enough to set both private and public RSA keys. However, having only n and e we can only set public RSA key.

Calculating d in OpenSSL:
```c++
BN_CTX *ctx = BN_CTX_new();
BN_CTX *ctx2 = BN_CTX_new();
BN_CTX_start(ctx);

BIGNUM *r0 = BN_CTX_get(ctx);
BIGNUM *r1 = BN_CTX_get(ctx);
BIGNUM *r2 = BN_CTX_get(ctx);

BN_sub(r1, m_rsa->p, BN_value_one());
BN_sub(r2, m_rsa->q, BN_value_one());
BN_mul(r0, r1, r2, ctx);

m_rsa->d = BN_mod_inverse(nullptr, m_rsa->e, r0, ctx2);

BN_CTX_end(ctx);
BN_CTX_free(ctx);
BN_CTX_free(ctx2);
```

Calculating d in GMP:
```c++
  // d = e^-1 mod (p - 1)(q - 1)
  mpz_t p_1, q_1, pq_1;

  mpz_init(p_1);
  mpz_init(q_1);
  mpz_init(pq_1);
  
  mpz_sub_ui(p_1, m_p, 1);
  mpz_sub_ui(q_1, m_q, 1);
  mpz_mul(pq_1, p_1, q_1);
  mpz_invert(m_d, m_e, pq_1);
```

PR #1026, opened 2019-04-14
urlencode stuff

divinity76:
for example spaces should be %20 encoded and the @ in `Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz` should be encoded to %40

but the real problem is the possibility that these stings may include the `&` character, which should be encoded as %25 (but wasn't prior to this commit)


iryont:
Looks fine to me :)

PR #1037, opened 2019-07-01
Improve mouse move event propagation

diath:
This changes the mouse move event propagation to only send the event to widgets whose rect contains the mouse position. Currently otclient will propagate the event to every single widget on the screen, even those that aren't really interested in the event and it becomes worse the more windows/widgets you have on the screen. In the most pathological case in my client that would be up to 1500~ events per mouse move with the current implementation, down to just about 8~ in the same scenario with the patch.

This is even worse in some std::deque implementations where thousands of allocations and dealloactions for each mouse move would grind the client to a halt (which is what actually prompted me to investigate the issue).

The best way to go about testing this would be running unpatched otclient then moving your mouse quickly around your character and watching the FPS counter drop, then opening as many windows as possible (such as the skills, battle, vip panels and containers), and watching the FPS drop even further, with this patch there should be barely any observable difference.

I'm not entirely sure why it wasn't done this way initially as the mouse press propagation works properly and the only reason I can think of is the scrollbar slider which should react even if you hold it and move the mouse outside of the scrollbar rect however this is trivially fixed by manually propagating the mouse move event to the currently pressed widget (if any) in the UIManager.

### DIFF
--- a/src/framework/ui/uiwidget.cpp
+++ b/src/framework/ui/uiwidget.cpp
@@ -1712,12 +1712,15 @@ bool UIWidget::propagateOnMouseEvent(const Point& mousePos, UIWidgetList& widget
 
 bool UIWidget::propagateOnMouseMove(const Point& mousePos, const Point& mouseMoved, UIWidgetList& widgetList)
 {
-    for(auto it = m_children.begin(); it != m_children.end(); ++it) {
-        const UIWidgetPtr& child = *it;
-        if(child->isExplicitlyVisible() && child->isExplicitlyEnabled())
-            child->propagateOnMouseMove(mousePos, mouseMoved, widgetList);
+    if(containsPaddingPoint(mousePos)) {
+        for(auto it = m_children.begin(); it != m_children.end(); ++it) {
+            const UIWidgetPtr& child = *it;
+            if(child->isExplicitlyVisible() && child->isExplicitlyEnabled() && child->containsPoint(mousePos))
+                child->propagateOnMouseMove(mousePos, mouseMoved, widgetList);
+
+            widgetList.push_back(static_self_cast<UIWidget>());
+        }
     }
 
-    widgetList.push_back(static_self_cast<UIWidget>());
     return true;
 }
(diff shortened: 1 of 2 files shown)


iryont:
One of the coolest improvement I've seen here for a long time. Congratulations :)

iryont:
There are some differences with the behavior though, e.g. tooltip does not longer follow your cursor once it appears. There might be more "issues" like that, but this is a good optimization nonetheless.

diath:
I'll try to figure out a way to fix the tooltips in spare time and submit a follow up patch, let me know if anyone can find any other unintended side effects of this change.

src/framework/ui/uiwidget.cpp
```diff
@@ -1712,12 +1712,15 @@ bool UIWidget::propagateOnMouseEvent(const Point& mousePos, UIWidgetList& widget
...
 bool UIWidget::propagateOnMouseMove(const Point& mousePos, const Point& mouseMoved, UIWidgetList& widgetList)
 {
-    for(auto it = m_children.begin(); it != m_children.end(); ++it) {
-        const UIWidgetPtr& child = *it;
-        if(child->isExplicitlyVisible() && child->isExplicitlyEnabled())
-            child->propagateOnMouseMove(mousePos, mouseMoved, widgetList);
+    if(containsPaddingPoint(mousePos)) {
+        for(auto it = m_children.begin(); it != m_children.end(); ++it) {
+            const UIWidgetPtr& child = *it;
+            if(child->isExplicitlyVisible() && child->isExplicitlyEnabled() && child->containsPoint(mousePos))
```
TheSumm: As far as I can see `child->containsPoint(mousePos)` is not needed here, because the subsequent call to `propagateOnMouseMove` will check `containsPaddingPoint`.

iryont: I've noticed this as well, but it seems like did copy it from propagateOnMouseEvent which does the exact same thing. I wonder why was padding rect chosen in that case.

PR #1038, opened 2019-07-01
Fix tooltip position updating

diath:
This fixes the bug I introduced in #1037. The tooltip position is now updated through the onMouseMove event in the root widget which always receives mouse updates.


diath:
Though perhaps a better approach would be to only connect the event when the tooltip is displayed, and disconnect it when it's hidden. Let me know if we should use that approach instead.

diath:
Actually it should, it makes very little sense to update the tooltip when it's not displayed, I've updated the PR.

iryont:
I like it, lets merge it and good job as always.

TheSumm:
Good job

PR #1039, opened 2019-07-06
Correcting text offset

slavidodo:
The correct definition of offset is to displace the text no matter what the alignment is.


iryont:
Good job.

PR #1042, opened 2019-09-27
Improved Docker support

f0rmiga:
- Fixed Dockerfile syntax.
- Consolidated dependencies installation.
- Removed unneeded dependencies.
- Unwrapped single-line commands into multi-lines for better caching and readability.
- Added `strip` step for the final otclient binary.
- Used multi-stage Docker building for better final image sizing.
- Pinned the base image to a specific hash to avoid undesired changes to the `latest` tag.
- Removed build.sh and run.sh (these files didn't make sense).
- Updated README.md to reflect the changes.


diath:
You could set [CMAKE_INSTALL_PREFIX ](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html) for the PhysFS project so that the install target installs the dynamic file in the desired location instead of moving the file manually. Also you could use [nproc ](http://man7.org/linux/man-pages/man1/nproc.1.html) instead of grepping procfs.

f0rmiga:
@diath Thanks for catching those. For the `mv`, it was plain copy-pasta leftover from the previous code. We actually don't need to change its location.

f0rmiga:
@diath Sorry for last-minute changes to CI. Let's see how it goes now.

iryont:
Good job.

midacts:
The docker build is just for the linux otclient, right?

diath:
That's correct.

PR #1044, opened 2019-10-10
Tidy up the source code

diath:
* Replaced push_back calls with emplace_back where applicable.
* Replaced `size() == 0` and `size() != 0` with `empty()` and `!empty()`.
* Replaced C style loops for range for loops where applicable.
* Fixed mismatching arg names between function declarations and definitions.
* Replaced `NULL` and `0` (in the context of pointers) with `nullptr`.
* Remove unnecessary calls to `string::c_str()` where applicable.
* Replaced deprecated C headers with proper C++ headers.
* Removed unnecessary null pointer checks when deleting pointers
(deleting a null pointer has no effect).
* Fixed a potential memory leak in apngloader.cpp file.
* Replaced unsafe `strcpy` with `strncpy` in the `demangle_name` function.


diath:
Please note that the CI job is failing because of an issue with libphysfs on Ubuntu and not due to these changes (as documented on our wiki), this is addressed in #1042.

iryont:
As always good job.

PR #1045, opened 2019-10-10
Add default copy constructors to Color and Position classes

diath:
GCC 9 adds a new diagnostic (-Wdeprecated-copy) that addresses the following:
>The generation of the implicitly-defined copy constructor is deprecated if T has a user-defined destructor or user-defined copy assignment operator. 

This results in dozens of warnings in all files that include the Color and Position classes as those classes supply user-defined copy assignment but no explicit copy constructor.

### DIFF
--- a/src/client/position.h
+++ b/src/client/position.h
@@ -36,6 +36,8 @@ public:
     Position() : x(65535), y(65535), z(255) { }
     Position(uint16 x, uint16 y, uint8 z) : x(x), y(y), z(z) { }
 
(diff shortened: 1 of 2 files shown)


diath:
Please note that the CI job is failing because of an issue with libphysfs on Ubuntu and not due to these changes (as documented on our wiki), this is addressed in #1042 which updates the travis job file.

iryont:
As always good job.

PR #1046, opened 2019-10-10
Fix use of deprecated PhysFS functions

diath:
This replaces the deprecated PhysFS functions with their non-deprecated variant. The code should ***not*** break for older PhysFS versions as the functions existed in prior versions.
* `PHYSFS_read` => `PHYSFS_readBytes`
* `PHYSFS_write` => `PHYSFS_writeBytes`
* `PHYSFS_getLastError` => `PHYSFS_getErrorByCode(PHYSFS_getLastErrorCode())`
* `PHYSFS_addToSearchPath` => `PHYSFS_mount`
* `PHYSFS_removeFromSearchPath` => `PHYSFS_unmount`
* `PHYSFS_isDirectory` => `PHYSFS_stat` (+ stat struct flag check)

This however doesn't remove the deprecated PHYSFS_getUserDir for the following two reasons:
* It seems like the library doesn't like the order in which things are initialized in otc and PHYSFS_getPerfDir may return a null pointer in some cases (would require further investigation).
* It changes the default write dir to a more sensible default (~/.local/share/\<app\> on Linux, %appdata%/\<org\>/\<app\> on Windows) but I'm not sure if we want to change that behavior (existing users would require a manual copy of the stored files though I think getPerfDir returns a more appropriate place to store these files).

### DIFF
--- a/src/framework/core/filestream.cpp
+++ b/src/framework/core/filestream.cpp
@@ -68,7 +68,7 @@ void FileStream::cache()
         PHYSFS_seek(m_fileHandle, 0);
         int size = PHYSFS_fileLength(m_fileHandle);
         m_data.resize(size);
-        if(PHYSFS_read(m_fileHandle, m_data.data(), size, 1) == -1)
+        if(PHYSFS_readBytes(m_fileHandle, m_data.data(), size) == -1)
             throwError("unable to read file data", true);
         PHYSFS_close(m_fileHandle);
         m_fileHandle = nullptr;
@@ -97,7 +97,7 @@ void FileStream::flush()
             if(!PHYSFS_seek(m_fileHandle, 0))
                 throwError("flush seek failed", true);
             uint len = m_data.size();
-            if(PHYSFS_write(m_fileHandle, m_data.data(), 1, len) != len)
+            if(PHYSFS_writeBytes(m_fileHandle, m_data.data(), len) != len)
                 throwError("flush write failed", true);
         }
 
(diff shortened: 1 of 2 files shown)


diath:
>The code should not break for older PhysFS versions as the functions existed in prior versions.

Which obviously only applies to minor versions as major version changes break API backwards compatibility. We should merge #1042 and then update the travis file to pull the physfs 3 version instead and make that the minimum required to build otclient as most mainstream distros ship with that version by default (I have verified with Arch, Gentoo, Fedora, openSUSE, Debian stable and newest release of Ubuntu), that also includes MS Windows in the [vcpkg ports](https://github.com/microsoft/vcpkg/blob/master/ports/physfs/portfile.cmake#L2).

iryont:
Good job. However, how about we move definition of ```isDirectory``` to:

```bool ResourceManager::directoryExists(const std::string& directoryName)```

And use it instead? Seems a bit unnecessary for it to be an outside function.

iryont:
As always good job.

iryont:
Ohh, I did merge this PR out of being in hurry. I do suggest to fix directoryExists the way I did describe though.

diath:
Absolutely, I was focused on the fact that the original code called PHYSFS_isDirectory in both places so I abstracted it away to avoid repetition but you're right that `ResourceManager::fileExists` can just piggyback off `ResourceManager::directoryExists`. I have resolved this in #1049.

PR #1048, opened 2019-10-11
Fix an issue with Connection class linkage

diath:
There seems to be an issue with static storage linkage in debug mode that may result in undefined references to these variables, this should address the issue. I also checked with [forgottenserver](https://github.com/otland/forgottenserver/blob/master/src/connection.h#L27-L28) to see how they had handled the breaking change in Boost a while ago and it seems to be done in the same way (likely for the same reason).

### DIFF
--- a/src/framework/net/connection.cpp
+++ b/src/framework/net/connection.cpp
@@ -101,7 +101,7 @@ void Connection::connect(const std::string& host, uint16 port, const std::functi
     m_resolver.async_resolve(query, std::bind(&Connection::onResolve, asConnection(), std::placeholders::_1, std::placeholders::_2));
 
     m_readTimer.cancel();
-    m_readTimer.expires_from_now(boost::posix_time::seconds(READ_TIMEOUT));
+    m_readTimer.expires_from_now(boost::posix_time::seconds(static_cast<uint32>(READ_TIMEOUT)));
     m_readTimer.async_wait(std::bind(&Connection::onTimeout, asConnection(), std::placeholders::_1));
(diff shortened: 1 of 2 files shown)


iryont:
Ahh yea, I remember I had an issue like this on OS X, but my solution was to move them all to enum which contains ```SEND_BUFFER_SIZE``` and ```RECV_BUFFER_SIZE``` already:

```c++
class Connection : public LuaObject
{
    typedef std::function<void(const boost::system::error_code&)> ErrorCallback;
    typedef std::function<void(uint8*, uint16)> RecvCallback;

    enum {
        READ_TIMEOUT = 30,
        WRITE_TIMEOUT = 30,
        SEND_BUFFER_SIZE = 65535,
        RECV_BUFFER_SIZE = 65535
    };
```

And simply cast them as long:

```c++
m_readTimer.expires_from_now(boost::posix_time::seconds((long)READ_TIMEOUT));
```

diath:
That could work though I'd prefer we use static_cast for new changes rather than C style casts but I wonder if a typed enum would work here getting rid of the need to cast the type altogether, I'd have to look into it.

iryont:
I can only imagine it should work just fine since we don't have anything else in the enum except for size of buffers.

diath:
It looks like we cannot use a typed enum because of a SFINAE failure since boost::is_integral<enum> fails for typed enums, I've updated the PR with casting the enum values instead.

iryont:
Interesting, Anyway, looks good.

PR #1052, opened 2019-10-14
Fix a possible missing null terminator in the demangle_name output buffer

diath:
In #1044 I had introduced a subtle bug where the demangled symbol name buffer could end up without a null terminator if the length of the demangled symbol name was equal or greater than the buffer size, this fixes the issue.


iryont:
@diath I did send a message to @edubart to add you as collaborator. It is pointless for me to merge your pull requests since you are perfectly capable of collaborating on your own.

PR #1057, opened 2019-11-20
Fixes for hard to find bugs

4drik:
Fixes to bugs that I found while using the client.

### DIFF
--- a/modules/game_battle/battle.lua
+++ b/modules/game_battle/battle.lua
@@ -249,25 +249,25 @@ end
 
 function doShowCreatureAtBattle(creature)
   if doCreatureFitFilters(creature) then
-	  local hidePlayers = hidePlayersButton:isChecked()
-	  local hideNPCs = hideNPCsButton:isChecked()
-	  local hideMonsters = hideMonstersButton:isChecked()
-	  local hideSkulls = hideSkullsButton:isChecked()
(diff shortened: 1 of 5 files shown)


modules/game_battle/battle.lua
```diff
@@ -244,26 +244,32 @@ function doCreatureFitFilters(creature)
...
+function doShowCreatureAtBattle(creature)
+  if doCreatureFitFilters(creature) then
+	  local hidePlayers = hidePlayersButton:isChecked()
+	  local hideNPCs = hideNPCsButton:isChecked()
+	  local hideMonsters = hideMonstersButton:isChecked()
+	  local hideSkulls = hideSkullsButton:isChecked()
+	  local hideParty = hidePartyButton:isChecked()
+
+	  if hidePlayers and creature:isPlayer() then
+		return false
```
Sickness-chan: indentation for all the return statements here too

modules/game_console/console.lua
```diff
@@ -882,7 +882,7 @@ function sendMessage(message, tab)
     channel = 0
   end
 
-  local findIni, findEnd, chatCommandInitial, chatCommandPrivate, chatCommandEnd, chatCommandMessage = message:find("([%*%@])(.+)([%*%@])(.*)")
```
diath: This is unused but it's needed for multireturn unpack.

PR #1059, opened 2019-12-13
focus first choice in modalWindow by default (QoL)

Zbizu:
selecting first modalwindow choice by default like in cip client


diath:
Isn't it what this bit does?
https://github.com/edubart/otclient/blob/82661e1b036cd603241feda2313dce88a01ef68f/modules/game_modaldialog/modaldialog.lua#L56

Zbizu:
obraz
apparently current option doesn't have any effect

diath:
You're right, I wonder why focusNextChild fails then but I guess that's unrelated to this PR.

Issue #1040, opened 2019-07-07
Corpse causing tile draw issue


theshibbies:
### Steps to reproduce (include any configuration/script required to reproduce)
1. Create 64x64 corpse (Cyclops or GS for example)
2. Move near another 64x64 item (fir tree for example) so that parts of the corpse will overlap onto parts of the existing item 
3. You will see in certain positions the corpse will be stacked on top of the item and your player will be drawn on top of the existing item as well

### Expected behaviour
Your character should always be drawn underneath the existing 64x64 item and in some instances the corpse should be down below the existing item

### Actual behaviour
Your character will be drawn above the existing 64x64 item and in some cases the corpse is drawn above the existing item where it should not be

### Environment
TFS 1.3 / Client Version 10.98

![sprites](https://user-images.githubusercontent.com/43316702/60763438-799b0900-a042-11e9-93a8-1e1836f0d14c.png)

-- Extra issue that I assume involves the same code which happens with items with top flag.
![51724187-2b60f300-202a-11e9-9899-39b8ce731ae0](https://user-images.githubusercontent.com/43316702/60763439-7c95f980-a042-11e9-9e92-cb0e81ab5075.gif)

The issue seems to mainly lie within the Tile::draw function within tile.cpp. However I cannot seem to figure out how to fix it. Any insight would be helpful.

Ochmar:
Cipsoft has same problems (well, they have different problems but I believe the root stays the same for both):
https://i.imgur.com/YppDbBp.png
Also, I noticed that this is kind of drawing problem while using stairs (or anything that has elevation). And this time cipsoft solved it, by... adding wall behind stairs so it won't bug.
https://i.imgur.com/1J5k3bN.png

iryont:
Well, in general it is not easy task to fix it. You can change the drawing order to horizontal from left to right and from top to bottom or an actual diagonal one (drawing from top-left to bottom-right going diagonally), however, despite the fact that those drawing orders do fix certain cases which were described here (e.g. the stairs in @Ochmar post), but they do bring other issues. The point is it's always a trade-off unless some more advanced ordering algorithm is used.

Corpses wider than 1x1 are yet another issue which do require re-drawing certain tiles, so e.g. character stays always on top of the corpse (despite it being drawn after the character). Personally I did remove this as it w as unnecessary waste of performance which didn't even bring expected results:

https://github.com/edubart/otclient/blob/master/src/client/tile.cpp#L108
[...]

PR #1070, opened 2020-02-03
Getting rid of redraw on lying corpse

Ochmar:
In https://github.com/edubart/otclient/issues/1040#issuecomment-581173589 we had a discussion if tile:draw() redrawing in case of lying corpse either is needed or not. We came to conclusion that there are way too many downsides of this resolution and therefore we opt against redrawing. Both performance-wise and gameplay-wise it is better without redrawing.


iryont:
Good job. Lets merge it.

PR #1083, opened 2020-04-26
Changes to the VIP List

4drik:
Fix that descriptions and icons do not mix after removing, adding VIPs and restarting the client.


diath:
There's too many conflicts to merge this, you will have to create a banch against the current master and apply the patch. Also can you describe the issue/how to reproduce it?

4drik:
The issue you can reproduce on servers without extended VIP List functions.
Add few VIPs, add to some of them a description, for example theirs vocation or name or whatever to remember.
Now remove some of them and reset client.
You will see now it's mixed up because you deleted some of VIPs and for this moment OTC is saving descriptions and icons for ID.

I have been testing the fix for months on narutostory.net server and the problem has been completely solved. Due to the fact that I am not a specialist in using repo - if you want, you can merge it in a correct way.

PR #1085, opened 2020-05-03
Big number separating

AscuasFunkeln:
Using commas separating for big number in skills window to make the numbers more readable.
Updated latest file version.

### DIFF
--- a/modules/gamelib/util.lua
+++ b/modules/gamelib/util.lua
@@ -9,6 +9,7 @@ function dirtostring(dir)
     end
   end
 end
+
 function comma_value(n)
 	local left,num,right = string.match(n,'^([^%d]*%d)(%d*)(.-)$')
 	return left..(num:reverse():gsub('(%d%d%d)','%1,'):reverse())..right
--- a/modules/gamelib/util.lua
+++ b/modules/gamelib/util.lua
@@ -9,3 +9,7 @@ function dirtostring(dir)
     end
   end
 end
+function comma_value(n)
+	local left,num,right = string.match(n,'^([^%d]*%d)(%d*)(.-)$')
+	return left..(num:reverse():gsub('(%d%d%d)','%1,'):reverse())..right
+end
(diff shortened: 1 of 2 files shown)


diath:
This function produces trailing `.0` for natural numbers, it probably shouldn't or it will look silly.

AscuasFunkeln:
I just make it for skills in my project and work fine, it dont return any `.0`
Yeah, better idea is moving it to make this useable in every modules.

modules/gamelib/util.lua
```diff
@@ -8,4 +8,8 @@ function dirtostring(dir)
       return k
     end
   end
-end
\ No newline at end of file
+end
```
diath, line 11: Newline after this end?

diath:
> I just make it for skills in my project and work fine, it dont return any .0

This is true in Lua 5.1 and Lua 5.2 but not Lua 5.3.

AscuasFunkeln:
> > I just make it for skills in my project and work fine, it dont return any .0
> 
> This is true in Lua 5.1 and Lua 5.2 but not Lua 5.3.

`return math.ceil(left..(num:reverse():gsub('(%d%d%d)','%1,'):reverse())..right)`
This will dont solve this trailing?

diath:
No, but I guess we can merge it as-is since it works fine with LuaJIT (our preferred Lua implementation) and Lua 5.1 which is the fallback in the CMake project when LuaJIT is not found.

scopz:
Isn't this behavior controller by the "tr" function?
I'm asking because locales should be taken into account when formatting numbers.
In Spanish (not sure about Portuguese) thousandSeparators are dots (.) and decimalSeparators (,) and this "tr" function checks this.
image

diath:
It looks like that was intended with the `tr` function but I don't think it's actually used for number values that much in modules, I guess all of them would have to be revised. Additionally, you would have to figure out what to do with locale strings that contain numbers on their own.

PR #1088, opened 2020-05-09
Spanish language revised

scopz:
Fixed some accents, typos and revised all the spanish translations.
I am from Spain, and I probably would have written some things differently, but I think they are understable for every Spanish speaker, even in Latam.


diath:
Do the Spanish characters render properly with our fonts? Also I don't speak Spanish to be able to verify the translation itself.

scopz:
I just took some pictures to check if the accents were correctly render.
eslocale
Also, after seeing it live, I made some more corrections, and fixed a bug on the other languages as well.

PR #1091, opened 2020-06-12
Slight fix of Tile::isLookPossible() function

4drik:
You should see through things that are translucent.

### DIFF
--- a/src/client/tile.cpp
+++ b/src/client/tile.cpp
@@ -527,7 +527,7 @@ bool Tile::isSingleDimension()
 bool Tile::isLookPossible()
 {
     for(const ThingPtr& thing : m_things)
-        if(thing->blockProjectile())
+        if(thing->blockProjectile() && !thing->isTranslucent())
             return false;
     return true;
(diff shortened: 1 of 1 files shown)


slavidodo:
I don't think so, I've checked the client and it only specifies the "unsight" flag which respects to block-projectile

4drik:
@slavidodo, you mean CipClient? In that case what function does the translucent option have?

slavidodo:
In Field:
```csharp
// in objects loop
CachedUnsight |= Object.IsUnsight; // BlockProjectile
```

In WorldMapStorage:
```csharp
public bool IsLookPossible(int x, int y, int z) {
	FieldRenderInfo info = GetFieldInfo(x, y, z);
	info.UpdateRenderCache();
	return !info.CachedUnsight;
}
```

4drik:
I understand, then I close the topic.
BTW. I'm not sure if translucent attribute works at all.
I had to create a new separate attribute to achieve my goal.

PR #1097, opened 2020-07-05
Patch - Mehah viewportcontrol

4drik:
A review of viewportcontrol by mehah

### DIFF
--- a/src/client/CMakeLists.txt
+++ b/src/client/CMakeLists.txt
@@ -76,6 +76,8 @@ set(client_SOURCES ${client_SOURCES}
     ${CMAKE_CURRENT_LIST_DIR}/towns.h
     ${CMAKE_CURRENT_LIST_DIR}/creatures.cpp
     ${CMAKE_CURRENT_LIST_DIR}/creatures.h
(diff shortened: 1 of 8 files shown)


diath:
Thanks for taking the time to make this change a proper PR, could you perhaps rename the files `viewmapcontrol` to `mapviewcontrol` and the class from `ViewportControl` to `MapViewControl`? This way we could keep it consistent with other map related classes.

iryont:
Personally I believe there are a lot of issues with this approach / code such as:

1) Aware range is static now defined by constant values.
2) ViewportControl class uses aware range while it should use drawing dimension of the MapView it belongs to.
3) ViewportControl class is not required, the basic idea (drawing less tiles depending on the walk direction) can be included within MapView class.
4) All of it should be done within ```updateVisibleTilesCache``` and not within drawing logic function which adds unnecessary indirection and logic to be calculated each frame.

To be honest all of it seems to be over-complicated. Please note this is my personal opinion.

4drik:
@diath Bakasura changes of PR:
https://github.com/edubart/otclient/pull/1098

PR #1100, opened 2020-07-09
fix lua files path

Raythan:
I don't know if this is specific for Debug module and x64 Platform. But was a correction that worked for me based on this post.
https://otland.net/threads/missing-2-src-files.264165/#post-2557282

### DIFF
--- a/src/framework/luaengine/lbitlib.cpp
+++ b/src/framework/luaengine/lbitlib.cpp
@@ -33,8 +33,8 @@
 #define LUA_LIB
 
 extern "C" {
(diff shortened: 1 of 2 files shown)


diath:
That just breaks building the client using CMake with LuaJIT on and everywhere with LuaJIT off.

Raythan:
Ok, I'll abandon the PR. Regards.

diath:
If this is what vcpkg does though I suppose a better way would be to update the include paths in the VS project instead.

PR #1102, opened 2020-07-13
resourcemanager:[bug fix] directoryExists method should resolvePath before checking

vfjpl:
https://github.com/otland/otclient/pull/43

Before directoryExists method called resolvePath on input. Later change was introduced that changed that behavior silently breaking some things(most notably CandyBot ui). This PR fixes that.


vfjpl:
@diath Any feedback?

vfjpl:
@diath feedback? also how to become maintainer of this project?

vfjpl:
The reason for this change is that `ResourceManager::fileExists` is not the only user of `ResourceManager::directoryExists` and previous change broke the behavior for different users of this functions.

PR #1105, opened 2020-08-23
change stdext::packet_vector to std::vector

vfjpl:
we don't use more memory by doing so but it can improve fps drop when many thing are happening on screen in the same time, because we reuse alocated memory insted of freeing it and alocating again.

packet_vector is copying data over on every push and erase because it always alocate new chunk of memory

### DIFF
--- a/src/client/tile.h
+++ b/src/client/tile.h
@@ -132,9 +132,9 @@ public:
 private:
     void checkTranslucentLight();
 
-    stdext::packed_vector<CreaturePtr> m_walkingCreatures;
-    stdext::packed_vector<EffectPtr> m_effects; // leave this outside m_things because it has no stackpos.
-    stdext::packed_vector<ThingPtr> m_things;
+    std::vector<CreaturePtr> m_walkingCreatures;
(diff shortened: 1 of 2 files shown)


vfjpl:
@andersonfaaria I'm cosiek#9731 on discord

@diath Well not really as I don't expect it to impact fps when nothing is changing. I used client with that modification and ram usage were normal 160-180mb and didn't saw any problems. So I thought I could get a feedback from you. But the benefits are that we are using more tested container. About vc and cmake files. I'm using linux and I don't think I'm the best person to modify vc files :)

diath:
Perhaps it would make more sense to keep `packed_vector` for tile things, as items are usually static objects for the most part so we can save the memory here and then use a regular `vector` for walking creatures and effects which tend to be highly dynamic, what do you think?

vfjpl:
I think that for small number of elements 1 to 4 vector allocates the same amout of memory as packet_vector. So I think it's better to change everything to just std::vector.

diath:
That seems right, the difference is negligible. I wanted to get rid of it completely in the same PR but couldn't update your PR because the PR branch/fork don't exist so I created another PR with your name as commit author: #1106.

PR #1106, opened 2020-08-26
Remove stdext::packed_vector

diath:


Kamenuvol:
I really didnt get what was the advantage of packed_vector. Could someone explain?
It was reallocating everything after each insert/erase? Why the hell did it exist at all?

diath:
A regular `std::vector` will use `std::allocator` (by default) that typically allocates memory in multiples of two but `stdext::packed_vector` would ensure that the underlying vector only keeps as much capacity as there are elements to reduce the memory usage (though it could also have been achieved by using custom `Allocator` for `std::vector<T, Allocator>` instead).
There was more discussion about it in #1105.

PR #1107, opened 2020-08-28
groundwork for more precise eventdispatcher

vfjpl:
This pr changes sheduledevent to use microsecond instead of millisecond. This in the future will allow more precise event timing.

first patch. small cleanup
second patch. changes scheduledevent to use microseconds and update eventdispatcher for new interface(it's the only direct user of scheduledevent)
third patch. change std::list to std::deque. This is better data structure for the job it's performing(fifo queue with push_front and push_back).

Later I/we can add new interface to eventdispatcher so we can schedule events with more time granularity. 

Please review. Also please take a look at this pr https://github.com/edubart/otclient/pull/1102

### DIFF
--- a/src/framework/core/eventdispatcher.cpp
+++ b/src/framework/core/eventdispatcher.cpp
@@ -74,13 +74,15 @@ void EventDispatcher::poll()
             event->execute();
         }
         m_pollEventsSize = m_eventList.size();
-        
+
         loops++;
     }
 }
 
 ScheduledEventPtr EventDispatcher::scheduleEvent(const std::function<void()>& callback, int delay)
 {
+    delay *= 1000;// SheduledEvent is in microseconds
+
     if(m_disabled)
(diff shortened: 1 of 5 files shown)


diath:
Since the dispatcher poll rate is tied to the application poll rate which in turn is tied to the FPS, does it even make sense to go for sub-millisecond precision?

vfjpl:
My thought was:
Given 60 FPS every frame is 16.(6) ms.
Given 120FPS every frame is 8.(3) ms.

let's take 10ms per frame as an example. 0.5(max error)/10 = 0.05 = 5% rounding error
the more fps the more the percentage goes up but also every single frame is less significant. So it's cancel the negative effect.
reverse is also true. The less fps then each frame is more significant but also the rounding error goes down.
is that significant enough? Don't know.

vfjpl:
OK. We will see if we need more precision from eventdispatcher after more work on walking animation. For now please merge first and third patch. Thanks :)

PR #1108, opened 2020-08-31
fix for laggy diagonal movement

vfjpl:


vfjpl:
@diath any feedback?

4drik:
There are only 2 active reviewers, @diath & @iryont, so we need to wait for them.

vfjpl:
@diath What do you think?

src/client/creature.cpp
```diff
@@ -579,20 +579,20 @@ void Creature::nextWalkUpdate()
...
+        }, getStepDuration(true) / 32);
     }
 }
 
 void Creature::updateWalk()
 {
-    float walkTicksPerPixel = getStepDuration(true) / 32;
-    int totalPixelsWalked = std::min<int>(m_walkTimer.ticksElapsed() / walkTicksPerPixel, 32.0f);
+    int stepDuration = getStepDuration(true);
+    int totalPixelsWalked = std::min<int>((m_walkTimer.ticksElapsed() * 32) / stepDuration, 32);
```
diath, line 589: This could result in division by zero when stepDuration is 0.

diath (changes requested):

src/client/creature.cpp
vfjpl: Division by zero was also there before :)
I'm not sure if it it better to set totalPixelsWalked to 0 or 32 when stepDuration is zero... As that mean that creature is infinity fast or not moving at all? I didn't encounter anything that returned 0 so I'm not sure.

diath: There was no division by zero before because it was dividing by a zero float which has a well defined behavior for it. I guess it could maybe happen with misconfigured monsters on the server side? I vaguely recall encountering something like this. I suppose we can just do something like this:
```cpp
    int duration = getStepDuration();
    int walkTicksPerPixel = std::max<int>(duration / Otc::TILE_PIXELS, 1);
    int totalPixelsWalked = std::min<int>(m_walkTimer.ticksElapsed() / walkTicksPerPixel, Otc::TILE_PIXELS);```

Comment (vfjpl) [reply src/client/creature.cpp:589]: Division by float zero and then casting that to int results in -2147483648 for me.
I don't know how it was behaving before.
Anyway If getStepDuration indeed return 0 then walk will be terminated in
```
    // terminate walk
    if(m_walking && m_walkTimer.ticksElapsed() >= getStepDuration())
        terminateWalk();
```
So maybe we should end walk earlier?
or I prefer something like this:
`    int totalPixelsWalked = stepDuration ? std::min<int>((m_walkTimer.ticksElapsed() * 32) / stepDuration, 32) : 0;`

diath:
I'd prefer if the `vector` to `deque` and timer changes were a separate PR. Also if you don't mind can we replace the `32`s in the movement code with `Otc::TILE_PIXELS`?

PR #1110, opened 2020-09-14
optional format string for stdext::date_time_string()

Source61:


src/framework/stdext/string.cpp
```diff
@@ -45,13 +45,13 @@ std::string resolve_path(const std::string& filePath, std::string sourcePath)
     return sourcePath + filePath;
 }
 
-std::string date_time_string()
+std::string date_time_string(const char* format)
```
diath, line 48: Can you change this to `const char* format/* = "%b %d %Y %H:%M:%S"*/`?

diath (changes requested):

src/framework/stdext/string.cpp
Source61: Done

PR #1121, opened 2021-01-04
Fix for Oldschool Tibia Light System (7.x)

okk3:
The oldschool Tibia formulas for Light Spells go as following: (Light: 6 light intensity, Great Light: 8 light intensitiy, Ultimate Light: 10 light intensity). Clearly by using 8 we're capping the distance where ambientLight is drawn.

Since this all is mostly sent by lua spells from TFS to the client, having optional full light (with a gamemaster for example) is just nice, as 10 only won't make the whole view range bright.

You can also limit this via modules if you don't want your players having

### DIFF
--- a/src/client/lightview.cpp
+++ b/src/client/lightview.cpp
@@ -28,6 +28,9 @@
 #include <framework/graphics/image.h>
 
 enum {
+#ifdef OLDSCHOOL_LIGHT
+    MAX_LIGHT_INTENSITY = 15,
+#endif
     MAX_LIGHT_INTENSITY = 8,
     MAX_AMBIENT_LIGHT_INTENSITY = 255
 };


src/client/lightview.cpp
```diff
@@ -28,6 +28,9 @@
 #include <framework/graphics/image.h>
 
 enum {
+#ifdef OLDSCHOOL_LIGHT
+    MAX_LIGHT_INTENSITY = 15,
+#endif
     MAX_LIGHT_INTENSITY = 8,
```
vfjpl, line 34: This will overwrite (or it won't compile) earlier MAX_LIGHT_INTENSITY

okk3: Sorry, I was high as hell yesterday, I'll correct this :)

peonso (commented):
I would straight change the value to 15, values come from the server, no idea why capping it client side. Or just make a protocol check.

PR #1123, opened 2021-01-10
Feature Clickable highlight words in npc channel

Ayato96:


okk3 (approved):

EgzoT:
I tested that and how it works:
* Left click on console text with highlighted words.
* Than shows menu with highlighted words.
* Choose word and then client will say this word on npc chat.

image

EgzoT (approved):

modules/game_console/console.lua
```diff
@@ -572,9 +574,29 @@ function addTabText(text, speaktype, tab, creatureName)
         -- Recalculate positions as braces are removed
         highlightData[(i-1)*3+1] = dataBlock._start - ((i-1) * 2)
         highlightData[(i-1)*3+2] = dataBlock._end - (1 + (i-1) * 2)
+	    highlightOptions[#highlightOptions + 1] = dataBlock.words
```
diath, line 577: Indentation.

modules/game_console/console.lua
```diff
@@ -572,9 +574,29 @@ function addTabText(text, speaktype, tab, creatureName)
...
       end
       label:setText(text)
 
+      if (#highlightOptions > 1) then
```
diath, line 581: No braces around if statements.

modules/game_console/console.lua
```diff
@@ -572,9 +574,29 @@ function addTabText(text, speaktype, tab, creatureName)
...
+              local menu = g_ui.createWidget('PopupMenu')
+
+              for k, v in pairs(highlightOptions) do
+                menu:addOption(v, function() sendMessage(v, consoleTabBar:getTab('NPCs')) end)
+              end
+              menu:display(mousePos)
+          end
+        end
+
+      elseif (#highlightOptions == 1) then
```
diath, line 593: No braces around if statements.

modules/game_console/console.lua
```diff
@@ -572,9 +574,29 @@ function addTabText(text, speaktype, tab, creatureName)
...
 
+      if (#highlightOptions > 1) then
+        labelHighlight.onMousePress = function(widget, mousePos, button)
+          if (button == MouseLeftButton) then
+              local menu = g_ui.createWidget('PopupMenu')
+
+              for k, v in pairs(highlightOptions) do
+                menu:addOption(v, function() sendMessage(v, consoleTabBar:getTab('NPCs')) end)
+              end
+              menu:display(mousePos)
```
diath, line 589: Add the following before displaying the menu to make sure the menu closes when the game ends:
```lua
menu:setGameMenu(true)
```

modules/game_console/console.lua
```diff
@@ -572,9 +574,29 @@ function addTabText(text, speaktype, tab, creatureName)
...
+          end
+        end
+
+      elseif (#highlightOptions == 1) then
+        labelHighlight.onMousePress = function(widget, mousePos, button)
+          if (button == MouseLeftButton) then
```
diath, line 595: No braces around if statements.

modules/game_console/console.lua
```diff
@@ -572,9 +574,29 @@ function addTabText(text, speaktype, tab, creatureName)
...
       label:setText(text)
 
+      if (#highlightOptions > 1) then
+        labelHighlight.onMousePress = function(widget, mousePos, button)
+          if (button == MouseLeftButton) then
+              local menu = g_ui.createWidget('PopupMenu')
+
+              for k, v in pairs(highlightOptions) do
+                menu:addOption(v, function() sendMessage(v, consoleTabBar:getTab('NPCs')) end)
+              end
```
diath, line 588: Rename `v` to `message` for clarity?
```lua
for _, message in pairs(highlightOptions) do
    menu:addOption(message, ...)
end
```

diath (commented):
Unfortunately I don't think that's what people expect from clickable word highlights, the feature should be able to detect clicks on individual keywords and send those instead of sending a popup menu with all the options. This probably could work by creating overlay text widget that would fill the original text label and then calculating the spacing to add margins for each overlay highlight, and handling mouse press events from there. Although it would probably require some more thought put into it with regards to text selection and text wrap.

EgzoT:
> Unfortunately I don't think that's what people expect from clickable word highlights, the feature should be able to detect clicks on individual keywords and send those instead of sending a popup menu with all the options. This probably could work by creating overlay text widget that would fill the original text label and then calculating the spacing to add margins for each overlay highlight, and handling mouse press events from there. Although it would probably require some more thought put into it with regards to text selection and text wrap.

I have similar feelings, but I approved changes because that's the solution give small advantages and... work. :thinking:

diath:
Superseded by #1148.

PR #1128, opened 2021-03-06
keep logs in deque and don't make unnecessary copy

vfjpl:


vfjpl:
some small improvements after that caught my attention. We don't need to make a copy because 'm_onLog' can't change anything. I think onLog is used when you bring up client console so lua can show you logs.

PR #1129, opened 2021-03-24
prepare bitmapfont for unicode support

vfjpl:


vfjpl:
@diath This PR cleans up bitmapfont module. Even if we don't succeed in adding unicode support, those changes are still valuable.
Please review

src/framework/graphics/bitmapfont.cpp
```diff
@@ -27,6 +27,14 @@
...
 #include <framework/otml/otml.h>
 
+namespace
+{
+    // for performance reasons we use statics vectors that are allocated on demand
+    CoordsBuffer s_coordsBuffer;
+    std::vector<Point> s_glyphsPositions(1);
+    std::vector<int> s_lineWidths(1);
+}
+
```
diath, line 37: Any reason these are moved from function scope to an anonymous namespace if they're not used across functions?

src/framework/graphics/bitmapfont.cpp
```diff
@@ -328,16 +334,18 @@ std::string BitmapFont::wrapText(const std::string& text, int maxWidth)
...
+                outText += line;
+                line.clear();
+            }
         }
 
-        line += word + " ";
+        line += word + ' ';
     }
 
+    line.pop_back();
```
diath, line 347: This could result in UB if `wrapText` is called in on an empty string, it should probably be:
```cpp
if (!line.empty() {
    line.pop_back();
}
```

diath (changes requested):

src/framework/graphics/bitmapfont.cpp
vfjpl: To avoid overhead of checking if we already initialized static variable every time the function is called. It's pretty small but the comment said that we use static for performance reasons so I thought we can improve it further a little bit :P
https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

src/framework/graphics/bitmapfont.cpp
vfjpl: I think we will get one empty string in `wordsSplit`.

http://boost.2283326.n4.nabble.com/Result-of-boost-split-against-empty-string-changed-in-post-1-45-release-td4631147.html
https://stackoverflow.com/questions/46540255/boostsplit-pushes-an-empty-string-to-the-vector-even-with-token-compress-on

src/framework/graphics/bitmapfont.cpp
diath: It looks like it's not really applicable in ST environment as rearranging the things produces the same assembly (at least in our case) however it seems like the C++ Guidelines say that anonymous namespaces should be the preferred method of dealing with things like that so I guess we can keep the change.

src/framework/graphics/bitmapfont.cpp
diath: Maybe in that case we could skip all the code if the result is an empty string anyway and just add the following at the top of the function:
```cpp
    if (text.empty()) {
        return {};
    }
```

vfjpl: wouldn't that disable NRVO?

vfjpl: That change would be pessimising normal "non empty string" to improve edge case.

I also don't like that we always add space and then remove it.

diath: That would just trade an NRVO for an RVO without needlessly executing the entire function. The overhead of checking for an empty string is insignificant although I don't think the current implementation poses any significant overhead either, I just thought it would be logically more correct to just handle an empty string instead of adding and removing the space.

vfjpl:
@diath ok you have to help me. Which one do you prefer?
``` c++
    // compose lines
    std::string line(words[0]);
    for(uint i = 1; i < words.size(); ++i)
    {
        std::string& word = words[i];

        line.push_back(' ');
        uint lineSize = line.size();
        line.append(word);

        if(calculateTextRectSize(line).width() > maxWidth)
        {
            line.resize(lineSize);
            line.back() = '\n';
            outText.append(line);
            line.assign(word);
        }
    }
    outText.append(line);
```
``` c++
    // compose lines
    std::string line;
    for(const auto& word: words)
    {
        auto lineSize = line.size();
        line.append(word);

        if(calculateTextRectSize(line).width() > maxWidth)
        {
            if(lineSize)
            {
                line.resize(lineSize);
                line.back() = '\n';
                outText.append(line);
                line.assign(word);
            }
        }

        line.push_back(' ');
    }
    line.pop_back();
    outText.append(line);
```

diath:
I'm not sure, the second one looks cleaner but I think you could merge both conditions and first check the `lineSize` so that it does not call `calculateTextRectSize` if not needed:
```cpp
if (lineSize > 0 && calculateTextRectSize(line).width() > maxWidth) {
	// ...
}
```

vfjpl:
lineSize > 0 only is there to make it so the first pass will do nothing. Otherwise we will get empty first line if first word size is still > than maxWidth. Which itself can only happen when we have space for less than two characters because we will always get at least char + '-' from breaking huge words above.

src/framework/graphics/bitmapfont.cpp
vfjpl: @diath How did you tested the assembly? for me it is different https://godbolt.org/z/4Mn7zMPrK

vfjpl:
@diath OK I have fixed the empty first line case. And also the adding and removing space that annoyed us bought. The line composing algorithm now build line in place instead on creating new string every loop iteration so it should be much faster.

diath (approved):
Looks fine, do you want me to merge this as-is?

vfjpl:
@diath Ok now you can merge. Apparently auto makes a copy if your function returns reference and if you don't specify that you want a ref… :O

PR #1131, opened 2021-03-26
Bump minimum required cmake version to 3.1

marmichalski:
Also removes ubuntu-clang workflow, FindCXX.cmake (replaces with native feature), introduces ugly stuff to compile Windows binary with cmake, fixes (cleanups?) some of the Find* modules.

Windows non-luajit build fails, because of lua 5.2. I could downgrade vcpkg to version with lua 5.1, at least I could try that, if you want 😎


src/framework/CMakeLists.txt
```diff
@@ -156,16 +151,17 @@ if(NOT CMAKE_BUILD_TYPE)
...
-set(WARNS_FLAGS "-Wall -Wextra -Wno-unused -Wno-unused-parameter -Wno-unused-result")
-set(CMAKE_CXX_FLAGS                   "${CMAKE_CXX_FLAGS} ${WARNS_FLAGS} ${ARCH_FLAGS} ${CPP2011_FLAGS} -pipe")
-set(CMAKE_CXX_FLAGS_COMPILESPEED      "-O0")
-set(CMAKE_CXX_FLAGS_DEBUG             "-O0 -g")
-set(CMAKE_CXX_FLAGS_RELWITHDEBINFO    "-O1 -g -fno-omit-frame-pointer")
-set(CMAKE_CXX_FLAGS_RELEASE           "-O2")
-set(CMAKE_CXX_FLAGS_PERFORMANCE       "-Ofast -march=native")
+set(GCC_FLAGS "-Wall -Wextra -Wno-unused -Wno-unused-parameter -Wno-unused-result -pipe")
+if (WIN32)
+    set(GCC_FLAGS "/bigobj")
```
marmichalski, line 157: I do hope that it's just temporary.

CMakeLists.txt
```diff
@@ -28,15 +28,19 @@ if(WIN32)
...
+                       ${CMAKE_CURRENT_SOURCE_DIR}/src/otcicon.rc)
     set(executable_SOURCES ${executable_SOURCES} otcicon.o)
 endif()
 
 add_definitions(-D"VERSION=\\"${VERSION}\\"")
 
 # add client executable
 add_executable(${PROJECT_NAME} ${framework_SOURCES} ${client_SOURCES} ${executable_SOURCES})
+
+set_target_properties(${PROJECT_NAME} PROPERTIES CXX_STANDARD 11)
```
diath, line 41: We should bump it to at least C++14 (or C++17 preferably).

src/framework/CMakeLists.txt
diath: Shouldn't that check for `MSVC` flag in addition to `WIN32`? Perhaps we should also rename it from `GCC_FLAGS` to something like `COMPILER_FLAGS`?

marmichalski: Yes, was thinking about it, but am planning to build the project from sln file on windows soon. So glued this piece of 💩 in a hurry. Will certainly add the MSVC check.

CMakeLists.txt
marmichalski: I've decided not to bump it as per https://github.com/edubart/otclient/pull/1111#issuecomment-695783454

😎

diath: That's no longer applicable as the primary concern was updating the provided SDK which is now considered a deprecated method of building the client of Windows. The library incompatibility is no longer a concern with vcpkg being the preferred method of fetching libraries on Windows.

PR #1133, opened 2021-03-29
Fix & enable by default static linking on macOS

marmichalski:


src/framework/cmake/FindLua.cmake
```diff
@@ -6,7 +6,7 @@
 
 FIND_PATH(LUA_INCLUDE_DIR NAMES lua.h PATH_SUFFIXES lua51 lua5-1 lua5.1 lua)
 FIND_LIBRARY(LUA_LIBRARY NAMES)
-SET(_LUA_STATIC_LIBS lua51.a lua5.1.a lua-5.1.a lua.a)
```
marmichalski, line 9: This was an erroneous change before.

src/framework/sound/declarations.h
```diff
@@ -27,8 +27,13 @@
 
 #define AL_LIBTYPE_STATIC
 
+#if defined(__APPLE__)
+#include <OpenAL/al.h>
```
marmichalski, line 31: I am not sure about this change. OpenAL is a native macOS framework and headers are available under OpenAL path. `openal-soft` installed by brew needs to be force-linked to overwrite it.

PR #1138, opened 2021-03-31
Fix the use of deprecated C++ functions

diath:


src/client/map.cpp
```diff
@@ -856,7 +858,7 @@ std::tuple<std::vector<Otc::Direction>, Otc::PathFindResult> Map::findPath(const
                 neighborNode->cost = cost;
                 neighborNode->totalCost = neighborNode->cost + neighborPos.distance(goalPos);
                 neighborNode->dir = walkDir;
-                searchList.push(std::make_pair(neighborNode, neighborNode->totalCost));
+                searchList.emplace(neighborNode, neighborNode->cost);
```
marmichalski, line 861: 💩

marmichalski (approved):
Other than inconsistent styling, LGTM.

PR #1139, opened 2021-03-31
keep logs in deque and don't make unnecessary copy

vfjpl:
Some small improvements after that caught my attention. We don't need to make a copy because 'm_onLog' can't change anything. I think onLog is used when you bring up client console so lua can show you logs.

### DIFF
--- a/src/framework/core/logger.cpp
+++ b/src/framework/core/logger.cpp
@@ -34,6 +34,12 @@
 
 Logger g_logger;
 
+namespace
+{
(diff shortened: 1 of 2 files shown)


diath:
NOTE: Looking at git blame and file history it looks like initially the copy was made to prevent modifying the queue from a task while the queue is still being processed, and then later a mutex was added when the logger class was expanded to prevent that from happening as well, so the copy is probably no longer needed.

PR #1141, opened 2021-03-31
bitmapfont: allow for more than 256 glyphs

vfjpl:
Second step in unicode support. This PR adds support for more glyhps in font.

We have two ways from now. Either we use utf8 encoding and then make function that manipulate strings utf8 aware, or we use something like std::wstring so one "char" can hold more codepoints. The upside is that we basically just use bigger font and std::wstring in places that we want unicode and that it. Also I thing that with current font system, 16bit codepoints is maximum that we can reasonably use. So no Chinese characters. Any thoughts? And what to focus on? Do we want to use unicode in interfaces or chat first? What encoding to use in network messages?

### DIFF
--- a/src/framework/graphics/bitmapfont.cpp
+++ b/src/framework/graphics/bitmapfont.cpp
@@ -49,9 +49,14 @@ void BitmapFont::load(const OTMLNodePtr& fontNode)
     // load font texture
     m_texture = g_textures.getTexture(textureFile);
     Size textureSize = m_texture->getSize();
+    int numHorizontalGlyphs = textureSize.width() / glyphSize.width();
+    int numVerticalGlyphs = textureSize.height() / glyphSize.height();
+    int numGlyphs = numHorizontalGlyphs * numVerticalGlyphs;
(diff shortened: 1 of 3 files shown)


diath:
You shouldn't really fixate on the BitmapFont class too much as it's not a proper way to handle Unicode and adding support for more than 256 glyphs with no proper way to set custom glyphs outside that range seems pointless. If you really want to implement Unicode then the primary focus should be TrueType font loading, text vertex array generating (using glyph data from a TT font), and rendering, and once we have a way to load TrueType fonts and render the text we can then worry about how and where to integrate it with the rest of the client.
UTF-8 should be the way to go anyway, and you wouldn't have to worry about NetworkMessage class then since if you wanted to send an UTF-8 string you would just add raw length prefix and a series of u8 values.

Here's one such implementation that could be used as a reference: https://github.com/rougier/freetype-gl

vfjpl:
@diath should this freetype font replace current BitmapFont or be added as another font to chose from?

diath:
We should probably support both initially.

PR #1142, opened 2021-04-01
fix lua if compiled without sound

vfjpl:
This patch fix lua crash if otclient is compiled without sound support


diath:
Could you change the if statements to look like:
```lua
if condition then
    -- code
end
```

Instead of
```lua
if condition then -- code end
```

PR #1147, opened 2021-04-15
Remove walking bounce

ralke23:
Hi! This is not a request/issue, it is more a suggestion. I wonder if is possible to remove the walking bounce. I searched for commits, and walk fixes, but I think this is directly related to graphics, like is done here https://github.com/OTCv8/otcv8-dev/commit/de4e6cad1ad3934f8e5e0dfb28b9e2fbfa6f9630.


vfjpl:
Have you tried both normal and dash walking?

ralke23:
> Have you tried both normal and dash walking?
> Which one do you think work better?

I will do a record of the bounce, with the two options and upload here.
This was my first pull request and I merged 300 commit by error, sorry :/

ralke23:
Sorry about this, the bounce bug was in my server sources
`uint32_t groundSpeed;
int32_t stepSpeed = getStepSpeed();

Item* ground = tile->getGround();
if (ground) {
    groundSpeed = Item::items[ground->getID()].speed;
    if (groundSpeed == 0) {
        groundSpeed = 150;
    }
} else {
    groundSpeed = 150;
}

double duration = std::floor(1000 * groundSpeed) / stepSpeed;
int64_t stepDuration = std::ceil(duration / 50) * 50;

const Monster* monster = getMonster();
if (monster && monster->isTargetNearby() && !monster->isFleeing() && !monster->getMaster()) {
    stepDuration *= 3;
}

return stepDuration;`

This fix the issue, for 8.6 protocol.
Regards!

Issue #1073, opened 2020-03-20
VS 2017+vcpkg compiling update


Fenrisus:
Hello there. It's kinda outdated right now.
If someone has issues with compiling at Visual Studio 2017 +vcpkg manager here the little things, that you should to fix.
Well, let's go.

Since version 1.1.0 OpenSSL have changed their library names from:
libeay32.dll -> libcrypto.dll
ssleay32.dll -> libssl.dll
**settings.props** file should looks like that:

```
..blahblahblah.
//find those 
    <OTCLIENT_LIBDEPS>
        glew32.lib;
        zlib.lib;
        libcrypto.lib; <----change libeay32.lib to libcrypto.lib
        physfs.lib;
        openal32.lib;
        lua51.lib;
        ogg.lib;
        vorbisfile.lib;
        vorbis.lib;
        opengl32.lib;
        dbghelp.lib;
        kernel32.lib;
        user32.lib;
        shell32.lib;
        gdi32.lib;
        advapi32.lib;
    </OTCLIENT_LIBDEPS>
    
    <OTCLIENT_LIBDEPS_D>
        glew32d.lib;
        zlibd.lib;
        libcrypto.lib; <----change libeay32.lib to libcrypto.lib
        physfs.lib;
        openal32.lib;
        lua51.lib;
        ogg.lib;
        vorbisfile.lib;
        vorbis.lib;
        opengl32.lib;
        dbghelp.lib;
        kernel32.lib;
        user32.lib;
        shell32.lib;
        gdi32.lib;
        advapi32.lib;
	</OTCLIENT_LIBDEPS_D>
```

Next statement is LUA;
around the code, instead of

```
#include "lua.h"
#include "lauxlib.h"
#include "lualib.h" 
#include <lua.hpp>
```

**use luajit**

```
#include "luajit/lua.h"
#include "luajit/lauxlib.h"
#include "luajit/lualib.h" 
#include <luajit/lua.hpp>
```

Done, it's compiles fine with Visual Studio 2017 and vcpkg.
Tested latest sources at 20/03/2020

deviljin112:
You are god's sent. I've spent 4 hours trying to compile OTClient with fails everywhere, both on windows and linux... This was literally the entire issue in errors!

Although, where is the LUA part? what file as I cant find it anywhere in VS

renancavalieri:
Thank you @Fenrisus, it worked with Visual Studio 2019 as well.

Felipe21:
[...]

PR #1153, opened 2021-06-12
Fix vcpkg compiling

TheSumm:
I would appreciate if someone else would test whether this fixes the vcpkg setup. It should work without any adjustments following the instruction at https://github.com/edubart/otclient/wiki/Compiling-on-Windows.

Related: #1073


EPuncker (approved):
before these changes:
2021-06-12 10_22_32-otclient - Microsoft Visual Studio

after these changes:
2021-06-12 10_31_21-otclient - Microsoft Visual Studio

TheSumm:
Thanks @EPuncker

PR #1154, opened 2021-06-13
[color] fix warning on gcc 11

vfjpl:


vfjpl:
What should I change? I also think about moving definition from .h to .cpp to speed up compile time. (color.h) is included in almost all translation units and this inline functions have to be parsed (compiled as weak symbol ?) every time then.

PR #1167, opened 2021-11-24
Make UIWidget::resizeToText() respect widget padding and text offset like text-auto-resize does

diath:


src/framework/ui/uiwidgettext.cpp
```diff
@@ -56,6 +56,15 @@ void UIWidget::updateText()
     m_textMustRecache = true;
 }
 
+void UIWidget::resizeToText()
+{
+    auto textSize = getTextSize();
+    textSize += Size(m_padding.left + m_padding.right, m_padding.top + m_padding.bottom);
+    textSize += m_textOffset.toSize();
+
+    setSize(textSize);
```
marmichalski, line 65:
```suggestion
void UIWidget::resizeToText()
{
    setSize(getTextSize() + Size(m_padding.left + m_padding.right, m_padding.top + m_padding.bottom) + m_textOffset.toSize());
```

Is the temp var needed?

diath: The temporary is used for the sake of readability, it will be optimized away anyway.

Issue #103, opened 2012-10-01
Character blinks when walking diagonal


hrsantiago:
Screenshot:
http://i1113.photobucket.com/albums/k503/conde22223/dasaslo.png

Must be walking in a mountain border, from northeast to southwest.

PR #1169, opened 2021-12-12
Fix minimap black tiles replacing color tiles

vfjpl:
Monsters moving just outside visible map region where causing black tiles to replace color tiles on minimap. I thing that was caused by new empty tiles created to hold information about moving monsters. `!newWalkingTile->isEmpty` fixed this. This was added  because https://github.com/edubart/otclient/issues/103 . I was not able to reproduce it and I'm not sure how the original update on empty tiles was supposed to fix it. It was some time ago so perhaps other changes fixed the other issue and introduced this one.

@edit I was able to reproduce issue/103 :(

please review

Thanks.

### DIFF
--- a/src/client/creature.cpp
+++ b/src/client/creature.cpp
@@ -557,7 +557,7 @@ void Creature::updateWalkingTile()
             newWalkingTile->addWalkingCreature(static_self_cast<Creature>());
 
             // recache visible tiles in map views
-            if(newWalkingTile->isEmpty())
+            if(!newWalkingTile->isEmpty())
                 g_map.notificateTileUpdate(newWalkingTile->getPosition());
         }
         m_walkingTile = newWalkingTile;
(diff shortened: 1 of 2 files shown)


diath:
I can still reproduce the issue with this patch applied:
image

vfjpl:
> I can still reproduce the issue with this patch applied: image

What tibia version have you tested? Mine was 8.60

Also this screenshot looks weird. How did you make it? 

Mine looks like
Przechwycenie obrazu ekranu_2021-12-12_15-39-59

Is there a difference between different tibia versions in minimap apperence?

vfjpl:
@marmichalski Can you help test this patch? :)

marmichalski:
> @marmichalski Can you help test this patch? :)

No, not really. 🤷‍♀️

PR #1172, opened 2022-02-04
Fixed some old typos.

zeeb92:
These typos has been around for a few years now, it's a commonly asked question on how to solve it so let's just add it to the master already.

### DIFF
--- a/src/framework/luaengine/lbitlib.cpp
+++ b/src/framework/luaengine/lbitlib.cpp
@@ -33,8 +33,8 @@
 #define LUA_LIB
 
 extern "C" {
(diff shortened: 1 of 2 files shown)


marmichalski:
Not a typo.

diath:
As discussed here <https://github.com/edubart/otclient/pull/1111#issuecomment-695783454>, this needs to be fixed in the VS solution and not in the codebase.

marmichalski:
@zeeb92 see if #1174 helps in your case. 🥲

zeeb92:
I see.
Alright, I'm closing it then.

PR #1174, opened 2022-02-04
Correctly include vcpkg luajit in visual studio

marmichalski:
Supersedes #1172


diath:
@zeeb92 Can you confirm that it fixes your issue?

PR #1183, opened 2022-05-24
move big functions from .h to .cpp

vfjpl:
Move definition from .h to .cpp to speed up compile time. (color.h) is included in almost all translation units and this inline functions have to be parsed (compiled as weak symbol ?) every time.


diath:
Usually inlining small functions that get called frequently is done for performance reasons to avoid function call overhead but looking at the client startup time difference (where these functions are used most frequently) with your changes there seems to be no significant impact so I guess it's fine.

PR #1186, opened 2022-06-16
Fix incomplete type from std::atomic (for linux build)

dudantas:


dudantas:
Opened by mistake, please ignore

PR #1210, opened 2023-08-12
Use float literals where applicable, fix -Wdouble-promotion

vfjpl:
Use float literals to perform more calculation in float instead of performing implicit conversion to double and back to float.

This fixes warnings with -Wdouble-promotion

### DIFF
--- a/src/client/animatedtext.cpp
+++ b/src/client/animatedtext.cpp
@@ -53,7 +53,7 @@ void AnimatedText::drawText(const Point& dest, const Rect& visibleRect)
 
     if(visibleRect.contains(rect)) {
         //TODO: cache into a framebuffer
-        float t0 = tf / 1.2;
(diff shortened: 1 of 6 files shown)


diath:
Do we enable `-Wdouble-promotion` by default? Aren't most modern CPUs implementing doubles in hardware? It looks like this change is backwards because it would cater to older hardware instead?

vfjpl:
All of this calculations was already done in floats. double literals just caused unnecessary conversion from float to double and back to float.
"It looks like this change is backwards because it would cater to older hardware instead?"
No, why is that? New hardware works with floats perfectly well. floats are plenty precise for our use.

diath:
>  No, why is that? New hardware works with floats perfectly well. floats are plenty precise for our use.

You're right, not sure what I was thinking here when making that comment.

